PR #21, opened 2013-02-18
Changed DEBUG log level to ROSDEBUG.

danepowell:
Using DEBUG as a log level is problematic for developers, because DEBUG is frequently defined as a symbol when debugging projects. I suggest changing the log level to ROSDEBUG to avoid this namespace conflict.


mikeferguson:
This looks OK, but could you please check over the examples (especially in rosserial_arduino) and just make sure if we have any examples with DEBUG they get updated as well.

danepowell:
It looks like the examples all make good use of the functions logdebug() etc... instead of using the DEBUG and other log level constants.

PR #26, opened 2013-02-19
Removed leftover .hgignore.

danepowell:


danepowell:
Ah, it looks like I'm behind the curve :)

PR #33, opened 2013-03-20
Groovy devel

PaulBouchier:
Fixes to examples directory organization for embeddedLinux, change Paul's email address, fix initial lost connection message on both Arduino & embeddedLinux

Paul

### DIFF
--- a/rosserial_embeddedlinux/CMakeLists.txt
+++ b/rosserial_embeddedlinux/CMakeLists.txt
@@ -8,6 +8,10 @@ install(DIRECTORY src/ros_lib
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/src
 )
 
+install(DIRECTORY examples
+  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/src
+)
+
 install(DIRECTORY ../rosserial_client/src/ros_lib
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/src
 )
--- a/rosserial_embeddedlinux/package.xml
+++ b/rosserial_embeddedlinux/package.xml
@@ -5,7 +5,7 @@
     Libraries and examples for ROSserial usage on Embedded Linux Enviroments
   </description>
   <author>Paul Bouchier</author>
-  <maintainer email="paul@fixme.com">Paul Bouchier</maintainer>
+  <maintainer email="paul.bouchier@gmail.com">Paul Bouchier</maintainer>
   <license>BSD</license>
   <url>http://ros.org/wiki/rosserial_embeddedlinux</url>
 
--- a/rosserial_embeddedlinux/src/rosserial_embeddedlinux/make_libraries.py
+++ b/rosserial_embeddedlinux/src/rosserial_embeddedlinux/make_libraries.py
@@ -85,7 +85,8 @@ rospack = rospkg.RosPack()
 # copy ros_lib stuff in
 rosserial_arduino_dir = rospack.get_path(THIS_PACKAGE)
 shutil.copytree(rosserial_arduino_dir+"/src/ros_lib", path+"/ros_lib")
+shutil.copytree(rosserial_arduino_dir+"/src/examples", path+"/examples")
 
 # generate messages
-rosserial_generate(rospack, path, ROS_TO_EMBEDDED_TYPES)
+rosserial_generate(rospack, path+"/ros_lib", ROS_TO_EMBEDDED_TYPES)
 
--- a/rosserial_python/src/rosserial_python/SerialClient.py
+++ b/rosserial_python/src/rosserial_python/SerialClient.py
@@ -344,6 +344,7 @@ class SerialClient:
 
         rospy.sleep(2.0) # TODO
         self.requestTopics()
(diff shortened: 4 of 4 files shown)


mikeferguson:
Manually merged -- this looks like the merge is all good -- but if you get a chance to test it, please do.

PaulBouchier:
Tested. Hello World has run for 24 hours on Arduino, no problem. It won't run long on embeddedLinux before dying. Still investigating. However, the initial "Lost sync" message is fixed.

Something broke make_libraries.py on both Arduino & embeddedLinux. They worked in an earlier version of the catkinized rosserial - maybe 0.4.2. The problem seems to be that in the previous, working version, roscd rosserial_arduino would take you into the catkin install directory, and that's where rosrun would run make_libraries.py from. The catkin_make install rule had deposited a properly built ros_lib directory there, so make_libraries could install it correctly. Now, roscd takes you to the catkin src directory, and I guess rosrun is running make_libraries from there, and ros_lib is incomplete there, so make_libraries fails.

I don't know how roscd knows where to teleport you to, but its changed behavior broke make_libraries.

PaulBouchier:
bouchier@Ubuntu-12:~/catkin_ws/src/rosserial/rosserial_arduino$ roscd rosserial_arduino
bouchier@Ubuntu-12:~/catkin_ws/src/rosserial/rosserial_arduino$ rm -rf ~/sketchbook/libraries/ros_lib
bouchier@Ubuntu-12:~/catkin_ws/src/rosserial/rosserial_arduino$ rosrun rosserial_arduino make_libraries.py /home/bouchier/sketchbook/libraries

Exporting to /home/bouchier/sketchbook/libraries
  File "/home/bouchier/catkin_ws/src/rosserial/rosserial_arduino/src/rosserial_arduino/make_libraries.py", line 88, in <module>
    rosserial_client_copy_files(rospack, rosserial_arduino_dir+"/src/ros_lib/")
  File "/home/bouchier/catkin_ws/src/rosserial/rosserial_client/src/rosserial_client/make_library.py", line 580, in rosserial_client_copy_files
    shutil.copy(mydir+"/src/ros_lib/"+f, path+f)
  File "/usr/lib/python2.7/shutil.py", line 117, in copy
    copyfile(src, dst)
  File "/usr/lib/python2.7/shutil.py", line 83, in copyfile
    with open(dst, 'wb') as fdst:
IOError: [Errno 2] No such file or directory: '/home/bouchier/catkin_ws/src/rosserial/rosserial_arduino/src/ros_lib/ros/duration.h'

Maybe this is caused by my ROS_PACKAGE_PATH:
ROS_PACKAGE_PATH=/home/bouchier/catkin_ws/src:/home/bouchier/catkin_ws/install/share
sending it to src first. Don't know where that's being set anymore, but will keep looking.

Also, it seems we're hitting this bug in python:
http://stackoverflow.com/questions/13193278/understand-python-threading-bug
which causes these errors on both Arduino & embeddedLinux
waiting for socket connection
Exception AttributeError: AttributeError("'_DummyThread' object has no attribute '_Thread__block'",) in <module 'threading' from '/usr/lib/python2.7/threading.pyc'> ignored
Exception AttributeError: AttributeError("'_DummyThread' object has no attribute '_Thread__block'",) in <module 'threading' from '/usr/lib/python2.7/threading.pyc'> ignored
Exception AttributeError: AttributeError("'_DummyThread' object has no attribute '_Thread__block'",) in <module 'threading' from '/usr/lib/python2.7/threading.pyc'> ignored
Exception AttributeError: AttributeError("'_DummyThread' object has no attribute '_Thread__block'",) in <module 'threading' from '/usr/lib/python2.7/threading.pyc'> ignored
Exception AttributeError: AttributeError("'_DummyThread' object has no attribute '_Thread__block'",) in <module 'threading' from '/usr/lib/python2.7/threading.pyc'> ignored
Exception AttributeError: AttributeError("'_DummyThread' object has no attribute '_Thread__block'",) in <module 'threading' from '/usr/lib/python2.7/threading.pyc'> ignored
Exception AttributeError: AttributeError("'_DummyThread' object has no attribute '_Thread__block'",) in <module 'threading' from '/usr/lib/python2.7/threading.pyc'> ignored
[INFO] [WallTime: 1363921553.287780] Note: publish buffer size is 512 bytes
[INFO] [WallTime: 1363921553.289761] Setup publisher on chatter [std_msgs/String]

Also, now that you've published rosserial debs, how should I describe to wiki readers which set of build instructions (catkin or rosbuild) to use if they installed deps? Should they roscd to rosserial_arduino & look at the version in package.xml & if it's higher than 0.4, use catkin? Or is there a more elegant approach?

mikeferguson:
RE: ros_lib directory -- 0.4.1 would not build on the farm. I had to change to pulling from the 2 different packages when running make_libraries -- although I apparently never pushed my last commit (and so it was still going off rosserial_arduino rather than rosserial_client as the directory sent to copy stuff). I just pushed an update -- mind testing that and letting me know if it works for you?

RE: Python threading thing -- is fixed upstream in python and ubuntu -- if you update your python debs you will probably see it go away.

RE: debs: well, they are broken for two reasons. 1) the ros_lib thing above, and 2) I just tried to run make_libraries and something I had not thought of was permission errors from shutil... I think we need to find a new way to move those files from debs.

PR #36, opened 2013-03-31
rosserial_embeddedlinux: Fix CMakeLists.txt ...

edsfocci:
so 'catkin_make install' can find 'examples' directory under 'src' directory in rosserial_embeddedlinux

This is so that the following part of the Arduino IDE Setup of rosserial can run properly:
http://ros.org/wiki/rosserial_arduino/Tutorials/Arduino%20IDE%20Setup#Installing_from_Source_onto_the_ROS_workstation


PaulBouchier:
Thanks for the update. I'd made it in my local tree but hadn't pushed it to the repo. Sorry for being slow, but happy to know you're using the package. Your change is merged into the repo now.
Paul

PR #46, opened 2013-04-18
Groovy devel

### LINKED ISSUE #43: Getting parameters before initing node?
http://answers.ros.org/question/60636/rosserial-parameter-loading-broken/

PaulBouchier:
Fix broken private parameters, which got broken when rosserial_embeddedLinux was merged in. Fixes #43 but breaks fork_server aspect of rosserial_embeddedLinux. Set fork_server to default to False for now


mikeferguson:
lgtm

PR #48, opened 2013-04-25
Rosserial_Modefied

tongtybj:
<a class="message" href="https://github.com/tongtybj/rosserial/compare/rosserial_bakui">
Add a checksum byte just after the message length bytes </a>for high frequency serial communication between ROS and Arduino.Tested on Arduino Atmega1280, publishing Int16MultiArray(8 \* 2 bytes) with 100Hz towards ROS.
Succeeded for consecutive publishing about 14 hour without any interruption.

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -44,12 +44,15 @@
 
 #define MODE_FIRST_FF       0
 #define MODE_SECOND_FF      1
-#define MODE_TOPIC_L        2   // waiting for topic id
-#define MODE_TOPIC_H        3
-#define MODE_SIZE_L         4   // waiting for message size
-#define MODE_SIZE_H         5
-#define MODE_MESSAGE        6
-#define MODE_CHECKSUM       7
+#define MODE_SIZE_L         2   
+#define MODE_SIZE_H         3
+#define MODE_SIZE_CHECKSUM  4   // checksum for msg size received from size L and H
+#define MODE_TOPIC_L        5   // waiting for topic id
+#define MODE_TOPIC_H        6
+#define MODE_MESSAGE        7
+#define MODE_MSG_CHECKSUM   8   // checksum for msg and topic id 
+
+
 
 #define MSG_TIMEOUT 20  //20 milliseconds to recieve all of message data
 
@@ -63,7 +66,6 @@ namespace ros {
       virtual int spinOnce()=0;
       virtual bool connected()=0;
     };
-
 }
 
 #include "publisher.h"
@@ -146,6 +148,7 @@ namespace ros {
        *  serial input and callbacks for subscribers.
        */
 
+
       virtual int spinOnce(){
 
         /* restart if timed out */
@@ -171,8 +174,8 @@ namespace ros {
           if( mode_ == MODE_MESSAGE ){        /* message data being recieved */
             message_in[index_++] = data;
             bytes_--;
-            if(bytes_ == 0)                   /* is message complete? if so, checksum */
-              mode_ = MODE_CHECKSUM;
+            if(bytes_ == 0)                  /* is message complete? if so, checksum */
+              mode_ = MODE_MSG_CHECKSUM;
           }else if( mode_ == MODE_FIRST_FF ){
             if(data == 0xff){
               mode_++;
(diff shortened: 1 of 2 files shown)


mikeferguson:
I took a quick glance at this, looks good, fixes a long-open ticket. However, I think we should tag one more 0.4.x release without this, and then put this into an 0.5.0/hydro branch since it is a pretty major change in the wire protocol and would probably really muck with anyone who does a debian update and suddenly the python node fails to work for them.

@PaulBouchier Does this have any implications on the embeddedlinux side?

PaulBouchier:
I looked over the changes and looks like they shouldn't have an impact that's specific to embeddedLinux. I'll try it out. I agree with a hydro introduction. I want to suggest that the wire protocol should have a version # embedded in it in some way, in order that mismatched ends will behave predictably - you'd want an older arduino to just not connect to a newer rosserial_python or visa versa. You wouldn't want it to connect and try to work and do weird things. I didn't see any protection against that. I suggest that the 0xff 0xff sync bytes could be changed to e.g. 0xff 0xfe to designate the new protocol. That would prevent the initial connection from ever happening between mismatched ends (it will happen - why should anyone expect that their Arduino which worked on Electric/Fuerte would suddenly start behaving strangely on Hydro).

A wire protocol version number would also permit protocol-sensitive interoperation, if someone felt that was worth developing.

PaulBouchier:
I've merged this pull request into the hydro branch and tested it on arduino and it seems to work fine when both rosserial_python and arduino have the matching change.

If arduino is at groovy and rosserial_python has this change, when rosserial_python starts, it immediately starts printing 
[ERROR] [WallTime: 1373105512.597335] Lost sync with device, restarting... 
This scenario is likely if someone updates to hydro on the linux box but doesn't also rebuild/reload the arduino code, and it's not a very helpful error message. People will likely think "rosserial is broken on hydro", rather than "I should rebuild my arduino".

I want to ask for comment on the proposal above: that the 2nd sync byte should be changed from 0xff to 0xfe, and that rosserial_python detect the mismatch and print an informative error message. 

If we are going to do that, this is the time, when we're changing the wire protocol.

Paul

mikeferguson:
I like that idea, especially since it prevents the possibility of an old-style packet where the first 3 bytes just happen to be a properly checksumed length

mikeferguson:
@mikepurvis do you have any thoughts on this?

mikepurvis:
It does seem an overdue change; I'm pretty sure this has affected me. It's too bad there isn't a way to add it that's backward compatible.

If we're breaking protocol compatibility anyway, it seems we should take advantage of the chance to add some other things. Is it too much bother to have a mini-review on the wiki of other suggested protocol changes? Another obvious thing would include explicit communication of a protocol version number from the firmware to host, so that the user is at least informed when it next breaks in the future.

PaulBouchier:
Mike - this is my post on the pull request that Fergs is referring to and

## asking you about:

If arduino is at groovy and rosserial_python has this change, when
rosserial_python starts, it immediately starts printing
[ERROR] [WallTime: 1373105512.597335] Lost sync with device, restarting...
This scenario is likely if someone updates to hydro on the linux box but
doesn't also rebuild/reload the arduino code, and it's not a very helpful
error message. People will likely think "rosserial is broken on hydro",
rather than "I should rebuild my arduino".

I want to ask for comment on the proposal above: that the 2nd sync byte
should be changed from 0xff to 0xfe, and that rosserial_python detect the
mismatch and print an informative error message.

If we are going to do that, this is the time, when we're changing the wire
protocol.

---

Effectively, changing the 2nd sync byte from 0xff to 0xfe encodes an
implicit protocol version # - if we do that we should define it to be
explicit: future protocol changes should use 0xfd, 0xfc, etc.

If there are other ideas, feel free to raise them, but I'm hopeful of
getting a debian into hydro, so we don't have much time. But if there's a
process for getting feedback, by all means let's use it.

Paul

mikepurvis:
Okay, yes. That sounds good to me. Implicit version coding in the lead bytes sounds like a good plan. Gives the option for the server to support multiple versions as appropriate, or at the very least to issue a helpful error message.

PaulBouchier:
Guys - I ran into a snag with the protocol version byte checking idea. I'd thought each side could check the other's protocol version, and if it detected a mismatch, do something appropriate. Arduino would return from SpinOnce() (as it currently does in case of header mismatch), and rosserial_python would print an error message. That simple implementation doesn't work because Arduino stays quiet until it gets a good packet; only then does it reply. So in the case of a hydro rosserial_python talking to a groovy Arduino, Arduino would never see a good packet (protocol byte would be wrong), so would never reply, so rosserial_python wouldn't get any message from Arduino to detect that there's a protocol mismatch, and eventually it times out & restarts - sends another packet to Arduino which Arduino throws away and nothing comes back. Well - it was a nice dream.

I thought about other approaches, such as having rosserial_python send an old-style packet initially, to trigger Arduino to respond with its packet (what a hack!), or making arduino accept either packet format and respond with whatever packet format it's built for (doesn't work for current groovy Arduino which would still throw away a hydro packet), or simply printing text as part of the "lost connection - resetting" message saying to check the version of the Arduino FW (simple, will work). Basically, it's hard to change the protocol by adding a byte in the middle of a packet and fool a groovy Arduino into sending back a packet - I'm about to give up. Does anyone have any good ideas for ways for hydro rosserial_python to detect it's talking to a groovy Arduino, while retaining the protection over the length that is the whole point of this exercise?

Paul

mikepurvis:
It would be possible to have the firmware side trigger the protocol upgrade.

Have the server lead with "classic" packets, and create a new magic service which the client can use to request the upgraded version. You could keep complexity in the client under control by just cooking up the bytestream for the request-upgrade message and baking it in, so you wouldn't be on the hook for multiple serialization schemes.

All this does have implications on the complexity of the server, though. Particularly troubling when there are multiple implementations of the server to keep up to snuff.

Maybe backward compatibility isn't worth all this—we could just make a clean break for hydro (or i-turtle), and have a full review of the message format to try to get it as right as possible. At Clearpath, we're using rosserial with both Grizzly and Kingfisher, so we have an interest in a clean field-upgrade story—but that can also be handled by just making firmware flashing a painless process for the end user.

mikepurvis:
On that note, what other schemes have been explored for dealing with the bit-error-in-length-field issue?

At the simplest, none of our rosserial firmwares use variable-length messages at all (outside of TopicInfo). There, there's perfect information on both sides about what to expect length-wise from messages.

For situations where there are variable-length messages in play, each receiver could go into a special long-message mode when a header declares, say, >50 bytes. Once it's in this mode of heightened suspicion, receive the thing in chunks, with an aggressive timeout between them. If the timeout expires, drop it on the floor and scan for the next 0xffff to resynchronize.

Another approach which is backward compatible protocolwise, but achieves the same end, would be to establish a magic length hinting topic, where either side could publish an advance notice of the length of the next-to-arrive variable length packet. The hint message could be as simple as a std_msgs/UInt16; it would be completely transparent to a user with Hydro on both sides, and either side being Groovy would just ignore the hints and carry on. The hints themselves are fixed-length and thus pose no risk, if the obvious measures are put in place to protect fixed-length packets. And then you get your checksummed length field without a breaking protocol change.

PaulBouchier:
Closing this pull request because the change discussed here has been merged into the pull request titled "Hydro Devel", along with some others. This pull request will not be merged into groovy because it introduces a wire protocol change which would potentially break stuff.

PR #53, opened 2013-07-05
Changed the build system to catkin for rosserial_xbee

tonybaltovski:
I have changed the build system for rosserial_xbee to catkin which as been placed in the metapackage rossserial_experimental for since ROS groovy.  I have tested this using a pair of series 1 xbees and it worked.  I think this would be a useful package for wireless serial communication.  Can someone else verify that it works as well? 

-Tony


mikeferguson:
@balt0 This pull request appears to include build directory, as well as a number of temporary files (CMakeLists.txt~ and setup.py~). Those have to go away before this can be merged.

mikeferguson:
@balt0 also a couple .pyc files.

mikeferguson:
I manually merged this in, without build folder or history of all those files.

tonybaltovski:
Thanks Mike  for merging this manually.  I started deleting the extra files but I could not delete the build directory from the website and I don't have access to my Linux machine.  Has anyone else tested this?

mikeferguson:
It doesn't "break" the build, but the install is definitely not quite right -- I'm working on that right now (but in the process found something else in my workspace has a busted install as well, so I'm fixing that first)

mikeferguson:
Ok, f9b71ab960abe2948e7e63cdb67b4068fdddb293 makes this installable by catkin (which means it will work on the buildfarm), I don't have an actual setup to test against, but it appears that all scripts are running, if you get a chance at some point, perhaps you can try it out on real hardware.

tonybaltovski:
Works like a charm.  Thank you.

PR #55, opened 2013-07-14
Hydro devel

PaulBouchier:
Introduced protocol version in 2nd byte of messages. 

@mikeferguson, please have a look at the change in node_handle.h to pre-enable the NEXT protocol change to be detected by rosserial_python: if client detects a protocol mismatch when configured_ is false, it sends a requestTimeSync() message. This will be ignored in the case of protocol mismatch, and seems it should be harmless if the protocol mismatch was due to data corruption and in fact there is not a protocol mismatch. 

@MikePurvis - thanks for suggestion to send a "trial message" using old protocol. I didn't go that route because the protocol is stateless at present - that would have made it stateful inasmuch as each side would have had to go from an uninitialized state to an initialized state  in sync with the other and I'm worried about a bunch of corner cases popping out of that where one side is initialized and the other not, and recovery from that. Seems like a can of worms. Instead I opted for the best messages I could emit to suggest that there might be a protocol mismatch in the case of non-response.

### DIFF
--- a/rosserial_python/src/rosserial_python/SerialClient.py
+++ b/rosserial_python/src/rosserial_python/SerialClient.py
@@ -324,7 +324,7 @@ class SerialClient:
 	# The protocol version is sent as the 2nd sync byte emitted by each end
         self.protocol_ver1 = '\xff'
         self.protocol_ver2 = '\xfe'
-	self.protocol_ver = self.protocol_ver2
+        self.protocol_ver = self.protocol_ver2
 
         self.publishers = dict()  # id:Publishers
         self.subscribers = dict() # topic:Subscriber
@@ -362,10 +362,10 @@ class SerialClient:
         data = ''
         while not rospy.is_shutdown():
             if (rospy.Time.now() - self.lastsync).to_sec() > (self.timeout * 3):
-	        if (self.synced == True):
-		    rospy.logerr("Lost sync with device, restarting...")
+                if (self.synced == True):
+                    rospy.logerr("Lost sync with device, restarting...")
 		else:
-		    rospy.logerr("Unable to sync with device; possible link problem or link software version mismatch such as hydro rosserial_python with groovy Arduino")
+                    rospy.logerr("Unable to sync with device; possible link problem or link software version mismatch such as hydro rosserial_python with groovy Arduino")
                 self.lastsync_lost = rospy.Time.now()
                 self.sendDiagnostics(diagnostic_msgs.msg.DiagnosticStatus.ERROR, "no sync with device")
                 self.requestTopics()
@@ -601,7 +601,7 @@ class SerialClient:
                 return -1
             else:
                 #modified frame : header(2 bytes) + msg_len(2 bytes) + msg_len_chk(1 byte) + topic_id(2 bytes) + msg(x bytes) + msg_topic_id_chk(1 byte)
-		# second byte of header is protocol version
+                # second byte of header is protocol version
                 msg_len_checksum = 255 - ( ((length&255) + (length>>8))%256 )
                 msg_checksum = 255 - ( ((topic&255) + (topic>>8) + sum([ord(x) for x in msg]))%256 )
                 data = "\xff" + self.protocol_ver  + chr(length&255) + chr(length>>8) + chr(msg_len_checksum) + chr(topic&255) + chr(topic>>8)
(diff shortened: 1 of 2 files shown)


PaulBouchier:
BTW, this commit incorporates tongtybj's pull request to groovy, which we decided to defer to hydro.

rosserial_python/src/rosserial_python/SerialClient.py
```diff
@@ -319,6 +320,12 @@ def __init__(self, port=None, baud=57600, timeout=5.0):
 
         time.sleep(0.1)           # Wait for ready (patch for Uno)
 
+        # hydro introduces protocol ver2 which must match node_handle.h
+	# The protocol version is sent as the 2nd sync byte emitted by each end
```
mikeferguson: You have a tab here -- there should be no tabs in our python code (as per ROS coding standard).

rosserial_python/src/rosserial_python/SerialClient.py
```diff
@@ -348,35 +355,32 @@ def requestTopics(self):
...
 
     def run(self):
         """ Forward recieved messages to appropriate publisher. """
         data = ''
         while not rospy.is_shutdown():
             if (rospy.Time.now() - self.lastsync).to_sec() > (self.timeout * 3):
-                rospy.logerr("Lost sync with device, restarting...")
+                if (self.synced == True):
+                    rospy.logerr("Lost sync with device, restarting...")
+		else:
```
mikeferguson: Tabs here as well.

rosserial_python/src/rosserial_python/SerialClient.py
```diff
@@ -348,35 +355,32 @@ def requestTopics(self):
...
                 continue
             flag[1] = self.port.read(1)
-            if ( flag[1] != '\xff'):
-                self.sendDiagnostics(diagnostic_msgs.msg.DiagnosticStatus.ERROR, "Failed Packet Flags")
-                rospy.loginfo("Failed Packet Flags ")
-                rospy.loginfo("Second Packet Header Failed, %s " %(flag[1]))
+            if ( flag[1] != self.protocol_ver):
+                self.sendDiagnostics(diagnostic_msgs.msg.DiagnosticStatus.ERROR, "Mismatched protocol version in packet: lost sync or rosserial_python is from different ros release than the rosserial client")
+                rospy.logerr("Mismatched protocol version in packet: lost sync or rosserial_python is from different ros release than the rosserial client")
+                rospy.loginfo("Protocol version was %s, expected %s " % (flag[1], self.protocol_ver))
```
mikeferguson: It might helpful to have a string output here, for instance, if it said "Protocol version was Rev. 0 (rosserial 0.4 and earlier), expected Rev. 1 (rosserial 0.5+)". This could be accomplished by having a little dictionary of protocol characters mapping to pretty print outputs describing them -- I think this would help a lot of users avoid confusion.

mikeferguson:
I added a couple of comments in the code. Generally looks OK to me -- I presume you've tested on embedded linux, how about Arduino?

PaulBouchier:
Fixed the tabs & changed my .exrc so it won't happen again. Added a dictionary as requested. All my testing has been done on Arduino with a 328 processor. I don't have an Atmel 168 handy, and my experience is it can't do much anyway - runs out of memory too fast. I re-checked and rosserial_embeddedlinux still runs fine, so I merged. I'll start the release procedure.

Since I merged tongtybj's work, I guess I should close his pull request?

mikeferguson:
If it works on a 328, should be fine on any of the regular arduinos, so I wouldn't worry about that.

And  yeah, close the pull request with a note to this one.

mikepurvis:
Just looking at this now—not sure I understand how it works. It looks like it only ever sends a new-style sync message in the requestTopics() call. How does the warning message get triggered?

PaulBouchier:
The warning message doesn't get triggered with a groovy Arduino, because groovy Arduino throws away the hydro packet. However, in the future, if we have another protocol change, say in i-ros, then a hydro-arduino will respond to a protocol mismatch by sending a request timesync, at which point i-ros rosserial_python will detect the protocol mismatch and issue the warning. I am putting in architecture to better support future protocol changes.

For groovy arduino, the change is to the timeout message (hydro rosserial_python times out when groovy Arduino doesn't respond). The timeout message suggests a couple of possible causes: possible link problem or mismatched SW versions.

PR #57, opened 2013-08-08
Several modification of rosserial for high speed commnunication 

tongtybj:
Now we use arduino as the basic controller for the attitude control of quadcopter. Meanwhile, we add a on-board compact computer on the quadcopter. In this system, we use rosserial to connect the arduino and computer. However, with the convetional rossreial version, we faced several troubles about serial communication, especially under the situation of high speed data transmission.

 The main problems are as follows:
1) Occurrence of Exception while reading from serial port
2) The overflow on the tx data buffer of arduino after sending SIGINT (Ctrl-c) to rosserial node
3) The duplicate registration of subscriber and publisher, while receiving TopicInfo for arduino

The first problem occurs when there are lots of routing processes within the arduino, such as the attitude controller module. We found that the process of rosserial often fell into halt, because of the exception while
calling the function of "self.port.read(x)" in the SerialClient.py. As we observed, the main exceptions were IOError 
and OSError. We thought that those exceptions would occur when the resource of serial port was temporarily unavailable. As a solution we add exception handling ("try" and "exception") to ignore those temporal errors which lead to the halt of rosserial node. After all, we consider that the exception handling is fairly necessary.

The second problem occurs when we interrupt the process of rosserial by using Ctrl-c, especially under the situation that the bps of tx data form arduino is quite high. We thought that if the bps of the tx data form arduino is quite fast, it  would fill out the buffer  before (c_time - last_sync_receive_time) > (SYNC_SECONDS*2200). In other word, the tx data buffer will fall  into overflow, before the arduino stops publishing message. As a result, the arduino will stop all of the serial communication, unless reboot the board.  Because we often use Ctrl-c to stop the process, we considered that signal handler is necessary. Put it concretely, we add an additional process which will send tx stop request to arduino after the node of rosserial receives SIGINT. At the side of arduino, the tx stop request can stop the arduino publishing data. To realize this point, we add the id of ID_TX_STOP to TopicInfo.msg.

The third problem occurs when the node of rosserial receives the response of self.requestTopics() from arduino for several times. The duplicate registration of publisher will lead to the multiple publishing of message from rosserial node to arduino, which is undesirable. Therefore, we add simple conditional expression before registraion. That is, if the publisher / subscriber has been already added to the lists.

Although most of those modifications are unrefined, they lead to the achievement of the high speed communication between arduino and computer. We hope those modification will help.


rosserial_msgs/msg/TopicInfo.msg
```diff
@@ -6,6 +6,7 @@ uint16 ID_SERVICE_CLIENT=4
 uint16 ID_PARAMETER_REQUEST=6
 uint16 ID_LOG=7
 uint16 ID_TIME=10
+uint16 ID_TX_STOP=11
```
mikeferguson: Adding this will cause the MD5 to change on TopicInfo -- which will break all existing installs. Personally I would say that such a change should not be merged into Hydro as we are well into "beta" for Hydro, and if someone has loaded firmware and then updates debs their system will fail to function.

mikeferguson: It also looks as though this isn't even used...?

rosserial_python/src/rosserial_python/SerialClient.py
```diff
@@ -455,10 +505,11 @@ def setupSubscriber(self, data):
...
             msg.deserialize(data)
-            sub = Subscriber(msg, self)
-            self.subscribers[msg.topic_name] = sub
-            self.setSubscribeSize(msg.buffer_size)
-            rospy.loginfo("Setup subscriber on %s [%s]" % (msg.topic_name, msg.message_type) )
+            if not msg.topic_name in self.subscribers.keys():
+                sub = Subscriber(msg, self)
+                self.subscribers[msg.topic_name] = sub
+                self.setSubscribeSize(msg.buffer_size)
+                rospy.loginfo("Setup subscriber on %s [%s]" % (msg.topic_name, msg.message_type) )
```
mikeferguson, line 503: So, if we do this, we don't ever update the subs/pubs -- say you update your firmware a bit, or plug in a different board, but keep the node running, it will fail in a non-obvious way. Perhaps there should also be a check for if the message is different, and then shutdown the old sub/pub and create a new one.

mikeferguson:
@PaulBouchier @mikepurvis I just noticed this PR, I've added some comments. Once the first one (the TopicInfo.msg change) is resolved, I think this could be merged -- I'd recommend one of you guys test it though as I have not.

rosserial_msgs/msg/TopicInfo.msg
tongtybj: Thank you for your replay. Though the change of MD5 as to TopicInfo is a big problem, TopicInfo included in ros_lib/ros/node_handle.h is converted from TopicInfo.msg. In other words, "ID_TX_STOP" is indirectly used in node_handle.h,

mikeferguson: So, what I would suggest for Hydro, is to make this line a comment, and then hardcode 11 into the node_handle.h, with the ID_TX_STOP as a comment. Then, for I-turtle, you can make this no longer a comment/hardcode match. (comments do not change the MD5)

tongtybj: Completely agree with your idea. I will implement this comment/hardcore match.

rosserial_python/src/rosserial_python/SerialClient.py
tongtybj, line 503: I will add additional code to the update problem. By the way, why the keys of subs and pubs are different, which subs use topic_name, while pubs use topic_id?

mikeferguson, line 503: I'm not actually sure there is a reason.

tongtybj:
Sorry for late to reflect the changes based on the discussion above.

Two main modification are done.
1. Add a comment/hardcore match for ID_TX_STOP, avoiding the change of the MD5 of TopicInfo.
2. Fix the potential problem of duplicated registration of same subscriber. 

As for 2, the core problem is that the instance of class "Subscriber" in SerialClient.py would register topic message as new subscriber , regardless of whether there is already a same registered topic message. The duplicated registration of subscribers call the same callback function again and again, which is undesirable. Therefore, we fix the setup  of subscriber to respond to  three kinds of topic messages.
  a) topic message with new topic name and new message type
    --> normal registration
  b) topic message with  existing topic name, but with different(new) message type.
    --> unsubscribe the existing topic message which has same topic name. substantially, register the new topic message  in the normal way. 
  c) topic message with existing topic name and same message type.
    --> do nothing.

rosserial_python/src/rosserial_python/SerialClient.py
```diff
@@ -455,10 +506,18 @@ def setupSubscriber(self, data):
         try:
             msg = TopicInfo()
             msg.deserialize(data)
-            sub = Subscriber(msg, self)
-            self.subscribers[msg.topic_name] = sub
-            self.setSubscribeSize(msg.buffer_size)
-            rospy.loginfo("Setup subscriber on %s [%s]" % (msg.topic_name, msg.message_type) )
+            if not msg.topic_name in self.subscribers.keys():
```
tongtybj, line 499: a) topic message with new topic name and new message type
--> normal registration

b) topic message with existing topic name, but with different(new) message type.
--> unsubscribe the existing topic message which has same topic name. substantially, register the new topic message in the normal way.

c) topic message with existing topic name and same message type.
--> do nothing.

tongtybj:
I have fixed the problems pointed out.
So can any one give me the feedback about those new modification?

mikepurvis:
This looks like a good change to me; would be fine to merge if you've tested it yourself.

It might be nice to wrap up that repeated try block in a small function to avoid all the repetition, but @PaulBouchier can make the call on that.

rflmota:
I'm having similar problems with yours @tongtybj ... I wonder if you could help me?

tongtybj:
@rflmota Yes, of course. So what is the concrete problem in your case.

By the way, I hope this pull request will be merged as soon as possible(maybe in ros-indigo).

rflmota:
Hi @tongtybj ,

First of all I'd like to than you for your attention.

Some of the issues that I'm getting are described in the following thread:
- https://github.com/ros-drivers/rosserial/issues/97

The other problem that I'm getting and this directely related to your merge request, specially #2 on your first post in this thread is that I'm running the following sketch in an Arduino Leonardo, but everytime I stop the node with 'ctrl+c' the next time I ran it, I need first to do an Hardware reset in the reset button of the Arduino board to get it working again or otherwise I keep getting the following error:

```
[ERROR] [WallTime: 1392416985.538923] Unable to sync with device; possible link problem or link software version mismatch such as hydro rosserial_python with groovy Arduino
```
- Arduino sketch :

``` c++
#define USE_USBCON
#include <ros.h>
#include <ros/time.h>
#include <imu_mpu6050/imu_raw.h>

#include <math.h>
#define _USE_MATH_DEFINES

#include "I2Cdev.h"
#include "MPU6050_6Axis_MotionApps20.h"

#if I2CDEV_IMPLEMENTATION == I2CDEV_ARDUINO_WIRE
    #include "Wire.h"
#endif

#define DEBUG TRUE

const char* pubName = "imu_raw";
const unsigned int pubRate = 10;

unsigned long pub_timer = 0;

MPU6050 mpu;

ros::NodeHandle nh;

imu_mpu6050::imu_raw inertial_msg;

ros::Publisher inertial_pub(pubName, &inertial_msg);

bool dmpReady = false;  // set true if DMP init was successful
uint8_t mpuIntStatus;   // hold actual interrupt status byte from MPU
uint8_t packetSize; // expected DMP packet size (default is 42 bytes)
uint8_t fifoCount;  // count of all bytes currently in FIFO
uint8_t fifoBuffer[64]; // FIFO storage buffer

// ================================================================
// ===               INTERRUPT DETECTION ROUTINE                ===
// ================================================================

volatile bool mpuInterrupt = false;     // indicates whether MPU interrupt pin has gone high
void dmpDataReady() {
    mpuInterrupt = true;
}

// ================================================================
// ===                      INITIAL SETUP                       ===
// ================================================================

void setup() 
{
    nh.getHardware()->setBaud(115200); //or what ever baud you want
    nh.initNode();
    //nh.spinOnce();
    nh.advertise(inertial_pub);
    //nh.spinOnce();

    bool devStatus;     // return status after each device operation (0 = success, !0 = error)

    // join I2C bus (I2Cdev library doesn't do this automatically)
#if I2CDEV_IMPLEMENTATION == I2CDEV_ARDUINO_WIRE
    Wire.begin();
    TWBR = 24; // 400kHz I2C clock (200kHz if CPU is 8MHz)
#elif I2CDEV_IMPLEMENTATION == I2CDEV_BUILTIN_FASTWIRE
    Fastwire::setup(400, true);
#endif

    // initialize device
    mpu.initialize();

    // load and configure the DMP
    devStatus = (bool) mpu.dmpInitialize();

    // supply your own gyro offsets here, scaled for min sensitivity
    mpu.setXGyroOffset(44);
    mpu.setYGyroOffset(-58);
    mpu.setZGyroOffset(35);
        mpu.setXAccelOffset(-1672);
        mpu.setYAccelOffset(-1264);
    mpu.setZAccelOffset(1382); // 1688 factory default for my test chip

    // make sure it worked (returns 0 if so)
    if (!devStatus ) {
        // turn on the DMP, now that it's ready
        mpu.setDMPEnabled(true);

        // enable Arduino interrupt detection
        //attachInterrupt(0, dmpDataReady, RISING); // interrupt 0 PIN2 do arduino UNO -> PIN3 do Arduino Leonardo
        //attachInterrupt(1, dmpDataReady, RISING); // interrupt 1 PIN3 do arduino UNO -> PIN2 do Arduino Leonardo
        attachInterrupt(2, dmpDataReady, RISING);   // interrupt 1 PIN0 do arduino Leonardo
        //attachInterrupt(3, dmpDataReady, RISING); // interrupt 1 PIN1 do arduino Leonardo
        //attachInterrupt(4, dmpDataReady, RISING); // interrupt 1 PIN7 do arduino Leonardo

        mpuIntStatus = mpu.getIntStatus();

        // set our DMP Ready flag so the main loop() function knows it's okay to use it
        dmpReady = true;

        // get expected DMP packet size for later comparison
        packetSize = mpu.dmpGetFIFOPacketSize();
    }
    else {
        // ERROR!
        // 1 = initial memory load failed
        // 2 = DMP configuration updates failed
        // (if it's going to break, usually the code will be 1)
    }
}

// ================================================================
// ===                    MAIN PROGRAM LOOP                     ===
// ================================================================

void loop() {
        nh.spinOnce();

        // if programming failed, don't try to do anything
        if (!dmpReady) return;

        // wait for MPU interrupt or extra packet(s) available
        while (!mpuInterrupt && fifoCount < packetSize);

        nh.spinOnce();

        // reset interrupt flag and get INT_STATUS byte
        mpuInterrupt = false;
        mpuIntStatus = mpu.getIntStatus();

        // get current FIFO count
        fifoCount = mpu.getFIFOCount();

        // check for overflow (this should never happen unless our code is too inefficient)
        if ((mpuIntStatus & 0x10) || fifoCount == 1024) {
            // reset so we can continue cleanly
            mpu.resetFIFO();

            // otherwise, check for DMP data ready interrupt (this should happen frequently)
        }
        else if (mpuIntStatus & 0x02) {
            // wait for correct available data length, should be a VERY short wait
            while (fifoCount < packetSize) fifoCount = mpu.getFIFOCount();

            nh.spinOnce();

            // read a packet from FIFO
            mpu.getFIFOBytes(fifoBuffer, packetSize);

            // track FIFO count here in case there is > 1 packet available
            // (this lets us immediately read more without waiting for an interrupt)
            fifoCount -= packetSize;

        }

        nh.spinOnce();

        // orientation/motion vars
        Quaternion q;           // [w, x, y, z]         quaternion container
        VectorInt16 aa;         // [x, y, z]            accel sensor measurements
        VectorInt16 aaReal;     // [x, y, z]            gravity-free accel sensor measurements
        VectorFloat gravity;    // [x, y, z]            gravity vector

        VectorInt16 gyro;

        mpu.dmpGetQuaternion(&q, fifoBuffer);
        inertial_msg.orientation.w = q.w;
        inertial_msg.orientation.x = q.x;
        inertial_msg.orientation.y = q.y;
        inertial_msg.orientation.z = q.z;

                mpu.dmpGetGyro(&gyro, fifoBuffer);
                inertial_msg.angular_velocity.x = gyro.x;
        inertial_msg.angular_velocity.y = gyro.y;
        inertial_msg.angular_velocity.z = gyro.z;

        mpu.dmpGetAccel(&aa, fifoBuffer);
        mpu.dmpGetGravity(&gravity, &q);
        mpu.dmpGetLinearAccel(&aaReal, &aa, &gravity);
        inertial_msg.linear_acceleration.x = aaReal.x;
        inertial_msg.linear_acceleration.y = aaReal.y;
        inertial_msg.linear_acceleration.z = aaReal.z;

                inertial_msg.stamp = nh.now();

        nh.spinOnce();

        inertial_pub.publish(&inertial_msg);

        nh.spinOnce();

}
```

mikepurvis:
Hey @tongtybj, I've just receive commit privs. This is going to need some manual attention to merge at this point, and in any case, I'd like to create a small function to eliminate the repeated try block in the Python server. We should also plan to modify rosserial_msgs to add the topic shutdown for Indigo, and I'd like to support that in the C++ `rosserial_server` as well.

Anyhow, all that to say, I definitely want this merged and am on the case. Sorry for the extremely long wait.

mikepurvis:
I've done a naive merge and added the tryRead function, here: https://github.com/ros-drivers/rosserial/tree/rosserialSpike

Needs to be tested with hardware prior to putting on the main branch, but I'm otherwise happy with how it looks. @tongtybj, can you review?

tongtybj:
Hi @mikepurvis ,
Thank you for your attention.

The try-block is quite refined compared to my commits.
And I found two bugs when I tested this branch on my code:
1. The number of bytes while reading the chk_byte
2. The indentation error around the definition of sendDiagnostics()
I sent a new pull request #102 , confirming that the new branch works fine. 

By the way, may I add the those modification on the C++ rosserial_sever ?

tongtybj:
Hi @rflmota ,
I am so sorry for the delay in my response.

Actually , the error message of "Unable to sync with device" often appears while the pc and arduino are disconnected. And this disconnection is often caused by the overflow of tx/rx buffer on arduino based on my experience.

Probably, after you stop the node with "ctrl-c", the serial connection on the PC side is shut-down;
however, the tx_stop_msg, which is the last message sent form pc to arduino, doesn't carry the correct tx_stop_id. and though the arduino receives the tx_stop_msg, it still keeps sending the messages to pc, leading to the overflow of buffer.

However, this is just my supposition. I hope you to check that whether the arduino receives the correct tx_stop_id and stop sending messages. Maybe a simple check function with blink of  LED 
is useful. If my supposition is right, please check "ros_lib" for sketch is latest version.

mikepurvis:
Hi @tongtybj, thanks for the review! I'll test this myself with hardware on Monday and then get it merged in—thanks again for your patience and persistence.

Yes, please feel free to add the same support to `rosserial_server`. If you don't get to it, I will add it myself in the coming weeks, along with parameter support. In tandem with this, I'm hoping to get [rosserial_test](https://github.com/clearpathrobotics/rosserial/tree/hydro-devel/rosserial_test) into a state where it provides some meaningful end-to-end coverage, which will make these sorts of enhancements more possible and verifiable in the future.

tongtybj:
Thank you for your quick response.
I have not used rosserial_server yet. So it will take some time for me to understand the codes, before I start to add the same modification.  I still want to try, but not sure if I can finish in the coming weeks. 
I did not notice the powerful tool, rosserial test, until I read your reply. I will try it in the near future.

k-okada:
Hello @tongtybj , we have quick question, When you say `high speed data transmission`, what the baudrate. Is 57600 high speed?

Cc: @iory

tongtybj:
Hello @k-okada, the high speed data transmission was operated at 115200. 57600 is a general and relatively low baudrate. Is there any problem with the data transmission?

k-okada:
@iory had trouble that the communication between the arduino board and computer stops after 10 minutes, and it partially resolved if he changed periodic rate of arduino program from 1000 to 1~10

Did you ever tested to connect more than one board to the PC? Or send multiple topics from the board to computer?

## 

◉ Kei Okada

2015/05/17 14:53、趙　漠居 notifications@github.com のメッセージ:

tongtybj:
The communication often stopped if too much packets accumulate in tx buffer
of the arduino board. But I think this kind of problem has been resloved.
And I have tested to connect multiple boards, sending multiple topics to
PC, and there is no any problem now.

Should I talk with @iory directly?

2015-05-17 15:05 GMT+09:00 Kei Okada notifications@github.com:

PR #58, opened 2013-08-18
Arduino Due compatibility

chuck-h:
A few modifications to make rosserial_arduino also compatible with Arduino Due (ARM3 cpu). See also http://wiki.jigren.org/index.php?title=Rosserial_On_Arduino_Due .
I did one regression check: pubsub example still builds and runs correctly on an Arduino UNO.


rosserial_arduino/src/ros_lib/ArduinoHardware.h
```diff
@@ -40,12 +40,18 @@
...
+#else
 #include <HardwareSerial.h>
+#define SERIAL_CLASS HardwareSerial
+#endif
 
 class ArduinoHardware {
   public:
     ArduinoHardware(HardwareSerial* io , long baud= 57600){
-      iostream = io;
+      iostream = static_cast<SERIAL_CLASS*>(io);
```
mikeferguson, line 54: This is really not the right way to do this. HardwareSerial on line 53 should become SERIAL_CLASS, this cast should go away. (I have no idea where it would find the definition of HardwareSerial when compiled on the DUE)

Have you tested this code on the DUE? ros/ros.h will actually call the default constructor (line 59), so I'm assuming that "UARTClass Serial" exists on the DUE?

rosserial_arduino/src/ros_lib/examples/Blink/Blink.pde
```diff
@@ -8,8 +8,10 @@
 
 ros::NodeHandle  nh;
 
+bool ledOn;
 void messageCb( const std_msgs::Empty& toggle_msg){
-  digitalWrite(13, HIGH-digitalRead(13));   // blink the led
```
mikeferguson, line 12: Does digitalRead() not exist on the due?

chuck-h:
In the bigger picture, this commit is all workarounds for imperfect Arduino code. Regarding HardwareSerial and its subclasses, I think the clean solution is for HardwareSerial to have all virtual member functions. Also I assume it's a bug that digitalRead() on an output port always returns LOW in Arduino Due.

I'm ok with a decision by ROS maintainters to simply wait until the Arduino code is fixed, rather than pushing workarounds.

mf> Have you tested this code on the DUE?

Yes, see link in original post.

mikeferguson:
A more generic version of the UARTClass fix was merged with #64, @PaulBouchier and I decided that we should not make the default Arduino sketches more complex, as Due will hopefully fix their implementation of digitalRead -- if not, a separate set of sketches should be created. This pull request is also now full of all sorts of unrelated commits -- so I'm closing it

Issue #37, opened 2013-04-03
C++ rosserial server


mikepurvis:
Potentially a good idea, for a variety of reasons—performance for one, also support of roscpp-only features such as the unreliable transport.

One possible implementation strategy would be building a "lib" package which provides the guts of the server, and then a script which generates the user's specific rosserial_server_xxx package, possibly even by scanning the firmware source and/or package dependencies to automatically determine which messages are necessary to include in the binary.

chuck-h:
I support this! I suspect the python rosserial driver is a bottleneck on Raspberry Pi. See http://wiki.jigsawrenaissance.org/ROS_on_RaspberryPi. Anything I can do to help?

mikepurvis:
I've got a very early prototype of this knocked together—minimal robustness, but it supports publishing and subscribing over sockets, and would probably be suitable for use with embeddedlinux. For anyone interested to take a look or play along at home: https://github.com/mikepurvis/rosserial/tree/hydro-devel/rosserial_server

PaulBouchier:
Closed by pull request #60

Issue #38, opened 2013-04-03
Missed subscriptions


mikepurvis:
We're having periodic issues where the MCU's request for a subscription fails to be processed by the rosserial serial node. It seems at present there is no mechanism in the protocol for an ack: setup is handled by a spew of TopicInfo messages, and ones with bit errors fall on the floor with a failed checksum.

My suggestion for how to change this would be to have a monitoring check in the firmware, which confirms that at least one message for each registered topic has been received, signalling a completed setup—if no message has been received, re-send the TopicInfo subscription command for that topic at some frequency, perhaps 1 Hz.

Similarly, in the server, the "Tried to publish before configured message" should probably trigger a re-sync attempt at some frequency, since that likely indicates a lost advertisement message.

mikepurvis:
The issues continue to affect some of our products at Clearpath. I'm adding the fixes proposed above to my C++ implementation of rosserial_server.

mikepurvis:
I ended up holding off on the proposed fixes, for two reasons:
- The logic is split: failed subscriptions must be detected in the client, whereas failed advertisements must be detected in the server.
- It increases complexity in the client.
- For occasional-use topics (like an estop trigger), there's a lot of chatter and repeated re-subscriptions.

A different proposal is to provide an optional "expect" parameter scheme with the server, so that the server can be started expecting to see certain advertisements and subscriptions from a connected client; if it doesn't seem them, then it can restart the connection until it does. This concentrates the robostness logic in the server, and makes it a fully optional, opt-in component.

mikepurvis:
Any objections to an expect/require parameter for rosserial_server? We're using rosserial-connected MCUs in Kingfisher and Grizzly, and it's increasingly becoming unacceptable to have the possibility of a missed setup message. Suggested API would be:

```
  <node pkg="rosserial_server" type="serial_node" name="rosserial_server">
    <param name="port" value="$(arg port)" />
    <rosparam param="require">
      publishers: [ 'sense' ]
      subscribers: [ 'has_wifi', 'cmd_drive', 'estop' ]
    </rosparam>
  </node>
```

When an MCU connects, it has 1000ms to create the declared publishers and subscribers—if it fails to do so, rosserial_server shuts down the connection and starts over. This should not ever be necessary for socket-connected MCUs, since TCP is guaranteed delivery; it would be solely for serial-connected clients.

[...]

PR #60, opened 2013-09-03
C++ rosserial_server

mikepurvis:
Working, with the following limitations:
- Functions standalone, but for full compatibility (eg, with rosbag), requires the message_info_service.py helper node to be running (see launch/socket.launch)
- No support (yet) for logging, parameters, or services. This won't be a priority for me/CPR, as we've typically just stuck to publish and subscribe.
- No tests at this point.

This implementation alternates between the VER1 and VER2 connection strings; it supports any groovy or hydro client. Socket and serial nodes are provided.

This PR addresses #37. Please let me know your thoughts.


mikepurvis:
@PaulBouchier, @mikeferguson, either of you guys want to take a look at this?

rosserial_server/package.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0"?>
+<package>
+  <name>rosserial_server</name>
+  <version>0.0.0</version>
+  <description>The rosserial_server package</description>
+
+  <!-- One maintainer tag required, multiple allowed, one person per tag -->
```
mikeferguson: I would remove all of these explanatory items from the package.xml -- no need to have them.

rosserial_python/nodes/message_info_service.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python
```
mikeferguson, line 1: I suggest adding a doc string to this to explain what it does in the general context of rosserial (mainly for future developers who come along and find it)

rosserial_server/package.xml
```diff
@@ -0,0 +1,50 @@
...
+  <!-- One maintainer tag required, multiple allowed, one person per tag --> 
+  <!-- Example:  -->
+  <!-- <maintainer email="jane.doe@example.com">Jane Doe</maintainer> -->
+  <maintainer email="mikepurvis@todo.todo">mikepurvis</maintainer>
+
+
+  <!-- One license tag required, multiple allowed, one license per tag -->
+  <!-- Commonly used license strings: -->
+  <!--   BSD, MIT, Boost Software License, GPLv2, GPLv3, LGPLv2.1, LGPLv3 -->
+  <license>TODO</license>
```
mikeferguson: License has to be filled in before this can be merged, and for clarity, should probably be BSD to match the rest of the rosserial.

rosserial_server/src/AsyncOkPoll.h
```diff
@@ -0,0 +1,42 @@
+#include <boost/bind.hpp>
+#include <boost/asio.hpp>
```
mikeferguson, line 35: Thumbs up for using boost::asio!

mikeferguson:
I only gave it a quick glance, didn't run anything, but given how standalone this is, it LGTM once the manifest is updated based on comments above. I'll leave it to @PaulBouchier to actually review and merge.

mikeferguson:
LGTM. @mikepurvis just a general note with github development: when adding commits to a pull request, no notification is actually sent to those following the pull request, so it's a good idea to add a comment afterwards (which will send a notification email).

PaulBouchier:
Mike, would you be willing to write some wiki documentation for this? I suggest it needs a page on the same level as rosserial_python, that explains the options and can be brief, but that some discussion of how to use it as an alternate to rosserial_python should go into one of the early tutorials for rosserial_arduino and rosserial_embeddedlinux.

rosserial_server/src/socket_node.cpp
```diff
@@ -0,0 +1,110 @@
+/**
```
PaulBouchier, line 1: Thanks for continuing to support sockets

mikepurvis, line 1: Definitely. We like sockets, some MCUs are a better fit on sockets anyway, and sockets are good for testing.

mikepurvis:
Re: Documentation—absolutely! Shall I do that now, or wait until it is merged and/or released?

Regarding ongoing development, although it's fully usable as-is there is obviously significant scope to improve on it as well, especially in supporting more of ROS and adding tests.

I also haven't tackled #38 in it to the degree which I would like. I will be making another PR at some point which implements [the scheme proposed here](https://github.com/ros-drivers/rosserial/issues/38#issuecomment-24238796).

PaulBouchier:
I suggest working on it now. That will help me test it. And as I go through
it I can find and fix doc bugs.

PaulBouchier:
Pulled in via pull #69

PaulBouchier:
Mike - can you look at the changes I've pulled into the repo to correct some compiler warnings about use of %ld when %d was what was needed. I hope there isn't some weird environment difference between us.

mikepurvis:
There's [some discussion here](http://stackoverflow.com/questions/1546789/clean-code-to-printf-size-t-in-c-or-nearest-equivalent-of-c99s-z-in-c) about how to printf a size_t in a cross-platform way.

I suspect that the easiest fix is to just eliminate printf-style log messages in favour of streams. I can create a PR for this.

PaulBouchier:
Oops - sorry for messing you up Mike, and delivering warnings to you. I got
the impression from the discussion you pointed to that %zu would work - if
you want to change that I'll merge it right away.

Paul

Issue #10, opened 2013-02-08
Unaligned access in TopicInfo serialize function on 32-bit platforms


danepowell:
I am trying to implement rosserial on an STM32F0 (ARM-based) board. When I try to connect to the board using serial_node.py, the code running on my board segfaults in rosserial_msgs/TopicInfo.h (which I generated using the latest version of make_library.py) at the memcpy functions. If I change my topic name to be 8 characters, I don't get the error. It seems that rosserial doesn't check that writes are block-aligned on 32-bit platforms.

danepowell:
Note that the serialization function in string.h also needs work.

danepowell:
Note that I've worked around this temporarily by [using protocol buffers](https://github.com/danep2/rosserial/tree/protobuf). The maintainer of rosserial has stated that he does not want to incorporate protocol buffers though, so a proper fix is still needed.

PaulBouchier:
I just realized I've seen this problem - thanks for reporting it Dane - I didn't know what caused it. The EmbeddedLinux version of rosserial was throwing alignment error messages to the console. It has to be this problem. It prevents use of rosserial on arm platforms, which is a growing problem - should be a problem on Arduino Due too. Needs to be fixed.

chuck-h:
I am seeing this bug as well on the Cypress PSoC4 micro (Arm Cortex-M0). 

BTW rosserial _does_ run on Arduino Due (e.g. https://github.com/ros-drivers/rosserial/pull/58), probably because its Cortex-M3 processor can do unaligned loads & stores (http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0552a/BABFAIGG.html)

mikeferguson:
M4 devices will also have trouble if gcc generates any instructions using the floating point unit.

lorenzoriano:
I'm having the same issue on an ARMv9. Can anybody point me to some resources where I can learn about this problem? I don't even know what to search for. What are alignment errors and how does one fix them?

Thanks!

chuck-h:
@Lorenzo

Basically an alignment error comes because most processors want big
datatypes (e.g. 4-byte, 32-bit integers) to be stored in memory at
addresses that are exact multiples of their size (e.g. addresses 1000,
1004, 1008 would be ok for 32-bit integers, 1001, 1002, 1003 would not).

A brief note on patching here:
[...]

PR #61, opened 2013-09-05
Fixed misalignment for 32 bit architectures

lorenzoriano:
This should fix issue issue #10. It removes the hard cast from char\* to uint32_t\* that causes misalignment errors on a 32 bit architecture (e.g ARM). I've tested it on a ARMv9 (Phidget SBC) and it seems to work fine.

### DIFF
--- a/rosserial_client/src/rosserial_client/make_library.py
+++ b/rosserial_client/src/rosserial_client/make_library.py
@@ -170,15 +170,16 @@ class StringDataType(PrimitiveDataType):
 
     def serialize(self, f):
         cn = self.name.replace("[","").replace("]","")
-        f.write('      uint32_t * length_%s = (uint32_t *)(outbuffer + offset);\n' % cn)
-        f.write('      *length_%s = strlen( (const char*) this->%s);\n' % (cn,self.name))
+        f.write('      size_t length_%s = strlen( (const char*) this->%s);\n' % (cn,self.name))
+        f.write('      memcpy(outbuffer + offset, &length_%s, sizeof(size_t));\n' % cn)        
         f.write('      offset += 4;\n')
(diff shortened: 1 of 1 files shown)


rosserial_client/src/rosserial_client/make_library.py
```diff
@@ -170,15 +170,16 @@ def make_declaration(self, f):
 
     def serialize(self, f):
         cn = self.name.replace("[","").replace("]","")
-        f.write('      uint32_t * length_%s = (uint32_t *)(outbuffer + offset);\n' % cn)
```
lorenzoriano, line 173: This is the offending line that causes misalignment on 32 bit architectures

rosserial_client/src/rosserial_client/make_library.py
```diff
@@ -170,15 +170,16 @@ def make_declaration(self, f):
...
+        f.write('      memcpy(outbuffer + offset, &length_%s, sizeof(size_t));\n' % cn)        
         f.write('      offset += 4;\n')
-        f.write('      memcpy(outbuffer + offset, this->%s, *length_%s);\n' % (self.name,cn))
-        f.write('      offset += *length_%s;\n' % cn)
+        f.write('      memcpy(outbuffer + offset, this->%s, length_%s);\n' % (self.name,cn))
+        f.write('      offset += length_%s;\n' % cn)
 
     def deserialize(self, f):
         cn = self.name.replace("[","").replace("]","")
-        f.write('      uint32_t length_%s = *(uint32_t *)(inbuffer + offset);\n' % cn)
```
lorenzoriano, line 181: This is the offending line that causes misalignment on 32 bit architectures

mikeferguson:
lgtm ... @chuck-h have you tried this version (I see you have your own)?

PR #62, opened 2013-09-05
fixed uninitialized arrays that would cause random segfaults

lorenzoriano:
Without setting to zeros the publishers and subscribers arrays, adding a new one (via e.g. advertise) would place it in the wrong position causing segfauilts later on (e.g. during spinOnce).

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -115,7 +115,12 @@ namespace ros {
        * Setup Functions
        */
     public:
-      NodeHandle_() : configured_(false) {}
+      NodeHandle_() : configured_(false) {
+        memset(publishers, 0, sizeof(publishers));
+        memset(subscribers, 0, sizeof(subscribers));
+        memset(message_in, 0, sizeof(message_in));
+        memset(message_out,0, sizeof(message_out));
+      }
       
       Hardware* getHardware(){
         return &hardware_;
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -120,6 +120,13 @@ namespace ros {
         memset(subscribers, 0, sizeof(subscribers));
         memset(message_in, 0, sizeof(message_in));
         memset(message_out,0, sizeof(message_out));
+        
+        req_param_resp.ints_length = 0;
+        req_param_resp.ints = NULL;
+        req_param_resp.floats_length = 0;
+        req_param_resp.floats = NULL;
+        req_param_resp.ints_length = 0;
+        req_param_resp.ints = NULL;
       }
       
       Hardware* getHardware(){


chuck-h:
I question whether this is necessary; and since I am implementing on a very small embedded system I try not to put in extra "belt & suspenders" code. As I understand it, the C/C++ language spec requires that all objects which are statically allocated at compile time (e.g. declared at file scope) get zero-initialized by cstartup. I'm using codesourcery arm-gcc 4.4.1 and the linker puts the nodehandle in the .bss section, so it's zeroed out for me. Are you actually seeing segfaults? Are you dynamically allocating the nodehandle object?

lorenzoriano:
I've seen segfaults on an ARMv9 processor and on an Intel machine, both with gcc (4.6 or 4.7). The initialization is only done once and it doesn't add any overhead, while it ensure the code works on a variety of platforms.

chuck-h:
@Lorenzo
Do you know whether the problems occurred with statically vs dynamically allocated nodehandle object? I suspect this factor more than a platform dependence.

Mostly asking to confirm/disconfirm my understanding of initialization. 

The initialization adds some program code space overhead. On a PC or RaspPi it's nothing but on an Arduino Uno, for example, every byte counts.

lorenzoriano:
In my tests it happened on nodehandle objects allocated on the stack, e.g. the main function. If somebody (like me) allocates the nodes in a function then they will incur in obscure segfaults that trace down to the non-initialized vectors.

I have no experience with Arduinos, does a single memcpy (or memset) incur in much overhead? As far as I know it's a highly optimized assembler routine (at least with gcc).

chuck-h:
Ok, it makes sense to me that if you declare the nodehandle inside a function (i.e. allocate it on the stack), it will be uninitialized. I am not sure why you do not want to follow the usage instructions at http://wiki.ros.org/rosserial/Overview/Initialization and shown in all the example code, however, which declares the nodehandle at file scope.

BTW the original Arduino boards (which seem to have motivated rosserial) have 32Kb of code space (2K of it lost to the bootloader) and 2Kb RAM. So rosserial was mostly written to be tight, tight, tight! It's a different world down here in embedded-land!

If the maintainers choose to keep rosserial super-tight (i.e. avoid the few dozen extra bytes of initialization code), I would ceertainly recommend explicit comments about the "statically-allocated" requirement, in addition to the "here's an example of how to do it" style of existing documentation.

lorenzoriano:
I see your point, and I really don't know if these modifications would have a significant impact on an Arduino build. For what I understand an Arduino program doesn't have a main function, therefore a global (hence static) NodeHandle is a must. But in the more generic case, i.e.  embedded systems less constrained than the Arduino, a proper initialization is necessary.

mikeferguson:
I think we can find a bit of a middle road here, by adding these memset calls, but wrapping them in a #ifndef SOME_ARDUINO_CONSTANT -- I'm not entirely sure what the constant is that would be defined on all AVR platforms.

lorenzo:
@chuck-h I would appreciate if @lorenzoriano were correctly tagged in messages in this project and not me :P

chuck-h:
@mikeferguson I suggest something more generic, e.g. #ifndef ROSSERIAL_COMPACT_CODE . I am porting to a very small ARM processor, not an AVR, and I would want this option. Furthermore, the bigger AVRs (e.g. AtMega256) can support a bit of code bloat in return for the additional flexibility lorenzoriano likes.

I like the fact that rosserial fits on a $2.00 processor, so it is becoming feasible to make things like wheel encoders and servos that talk directly to ROS.

PaulBouchier:
I'm concerned that I don't understand the operation of this code well enough to understand why this fix is complete, or whether 0 is a sentinal value somehow. 0 can cause a segfault if dereferenced as a pointer. 

For a segfault to occur when the array is uninitialized implies the values in the array are being used before initialization. How does that happen? in spinOnce() line 252 for example the subscribers array is indexed by the topic ID in the incoming message. Does the topic ID refer to a non-existent subscriber, or how does the segfault get triggered?

I do see that if node-handle is declared as a local object (on the stack), which is reasonable to do, any class instance variables will have uninitialized contents, and appropriate initialization should be done, or else dynamic allocation must be declared unsupported (undesirable). I'm trying to make sure we do the right initialization, of only the variables that matter. initializing arrays isn't usually needed unless the values in the arrays are used in a way that makes a 0 value special.

Thank you for your contribution Lorenzo. Can you help me understand better how the uninitiaized value gets used before being set by something like a subscription request.

lorenzoriano:
in file node_handle.h, function advertise (roughly line 320):

``` c++
/* Register a new publisher */    
bool advertise(Publisher & p)
{
for(int i = 0; i < MAX_PUBLISHERS; i++){
          if(publishers[i] == 0){ // empty slot
            publishers[i] = &p;
            p.id_ = i+100+MAX_SUBSCRIBERS;
            p.nh_ = this;
            return true;
          }
}
return false;
}
```

if the array `publishers` is not initialized to have all zeros, the first entries are considered to be pointing to a `Subscriber_*` object, only because the memory address is not 0. Initializing `publishers` fixes this problem, by making the check   `if(publishers[i] == 0)` correct.

The same goes for the other arrays.

rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -115,7 +115,19 @@
        * Setup Functions
        */
     public:
-      NodeHandle_() : configured_(false) {}
+      NodeHandle_() : configured_(false) {
+        memset(publishers, 0, sizeof(publishers));
```
PaulBouchier, line 119: Lorenzo, if you could replace the calls to memcpy with a for-loop that will address Mike's concerns about linking with memcpy.
for (int i=0, ptr=beginAddress; i<sizeof(struct); i++,ptr++) *ptr=0; 
I'll merge & do a release this weekend.

lorenzoriano, line 119: Paul, I'll do it next week as I'm off for the weekend

PaulBouchier, line 119: Actually, don't worry about this. I did it this morning, and I need to do a
release this weekend, to support MikePurvis. It will be checked into the
repo shortly.

Paul

lorenzoriano, line 119: Ok, thanks!

Issue #49, opened 2013-04-28
Serial Assignment Error When using Arduino Leonardo and IDE 1.04


futureneer:
Hello, I encountered the following error using the 1.04 IDE and a Leonardo when compiling my code against the ros_lib headers:

```
/home/kel/sketchbook/libraries/ros_lib/ArduinoHardware.h:53: error: cannot convert ‘Serial_*’ to ‘HardwareSerial*’ in assignment
```

This was fixed by changing Line 53 of ArduinoHardware.h from 

```
iostream = &Serial;
```

to

```
iostream = &Serial1;
```

as per these instructions:
http://arduino.cc/forum/index.php?topic=111354.0

I am not sure if this is specific to the leonardo, or the 1.04 IDE.  It may have to do with the fact that newer arduinos like the Leonardo have a virtual serial port (more info at: http://arduino.cc/en/Guide/ArduinoLeonardo#toc11).  Just wanted you to know that this is an issue, I am not sure right now how it could be resolved.  Perhaps something like in the above link where the type of arduino is detected and the iostream is assigned a different serial value for each.  This seems kind of messy, but it could work.

I really appreciate your work, it is wonderful to be able to use Arduino on ROS.

Cheers and thanks!
Kel Guerin

superxingzheng:
I solved the same problem by changing the 47th line of ArduinoHardware.h into 
ArduinoHardware(Serial_\* io , long baud= 57600){
and the 80th line into 
Serial_\* iostream;
I think this is the right way to solve this bug. changing into &Serial1 will cause future communication failure.
My method is tested with ROS Fuerte under Ubuntu 12.04 32bit and Arduino Leonardo.

mikeferguson:
I presume there is some sort of #define that tells you that you are compiling for a Leonardo -- if anyone wants to find out what that is and add a #ifdef #else #endif block, and submit a pull request with a patch for this, it'd be much appreciated and certainly merged.

PaulBouchier:
Kel, Zheng,

Thank you for reporting this issue. Can you help me understand your system configuration.

[...]

PR #64, opened 2013-09-11
add support for leonardo and due

mikeferguson:


mikeferguson:
@PaulBouchier this should address 
- https://github.com/ros-drivers/rosserial/issues/49
- https://github.com/ros-drivers/rosserial/pull/58 (at least, the compiling, examples using digitalread will still not work)

mikepurvis:
Not sure how much of interest this is, but we've been using rosserial with 32u4s for a while. Unfortunately, the source in question is proprietary, but I've extracted a minimal demonstration example and thrown it up on github.

Please check it out here: https://github.com/mikepurvis/rosserial_leonardo_example

Three catkin targets get created:

```
$ catkin_make help
...
... rosserial_leonardo_example_firmware
... rosserial_leonardo_example_firmware_upload
... rosserial_leonardo_example_ros_lib
```

Something like this could be the basis of a new rosserial_arduino package. An obvious next step would to wire up the install targets so that the compiled firmware is installed, along with the appropriate flashing utility.

PR #68, opened 2013-09-17
De-registration of subscribers and service clients.

lorenzoriano:
This prevents callbacks to be called even when a client program
terminates a connection.

When a client dies, the rospy.Subscriber connection is still alive. If a new client connects to the python node then it will receive 2 messages, and if it dies again the next time it will be 3, and so on..

With these modifications whenever a client dies all the connections are automatically destroyed, so a new client will receive only the right messages.


mikepurvis:
+1

PR #70, opened 2013-09-18
Inilining all the functions to allow compilation of files with multiple includes

lorenzoriano:
Right now if you try to compile together two files a.cpp and b.cpp, both of which `#include <ros.h>`, the linking phase will fail given that all the functions in the cpp files are defined twice. The reason is that some header files include cpp files, a choice that I fail to understand.

Ideally one should have everything in header files, but this fix allows cpp files to exist and a library to be created (as in the [wiki page](http://wiki.ros.org/rosserial_embeddedlinux/Tutorials/Advanced%20Techniques#Building_rosserial_embeddedlinux_as_a_library) ).

**Note for code size**: usually inlining code increases the code size (not necessarily!). However, considering that most of the rosserial library is already header only, therefore inlined by default, the additional functions should not make a substantial difference.

**Note for additional commit**: The file node_handle.h is modified by a different fix which is already the subject of pull request #62.

### DIFF
--- a/rosserial_client/src/ros_lib/duration.cpp
+++ b/rosserial_client/src/ros_lib/duration.cpp
@@ -37,7 +37,7 @@
 
 namespace ros
 {
-  void normalizeSecNSecSigned(long &sec, long &nsec)
+  inline void normalizeSecNSecSigned(long &sec, long &nsec)
   {
     long nsec_part = nsec;
     long sec_part = sec;
@@ -56,7 +56,7 @@ namespace ros
     nsec = nsec_part;
   }
 
-  Duration& Duration::operator+=(const Duration &rhs)
+  inline Duration& Duration::operator+=(const Duration &rhs)
   {
     sec += rhs.sec;
     nsec += rhs.nsec;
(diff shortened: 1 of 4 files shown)


PaulBouchier:
Adding @mikeferguson with correct spelling.

mikepurvis:
The choice to #include those source files is made by the client library, as that's where ros.h comes from. If you don't like that it's that way in the provided arduino or embedded libs, the easiest solution is to roll your own platform adapter (and ros.h), where your build system handles compiling and linking them in.

lorenzoriano:
I've done some research, and apparently gcc gladly inlines all the "simple" functions it can find in a object file (i.e. a source file plus all the recursively included files). This means that if one writes a source that includes ros.h, it is including (and probably inlining) all the functions defined in the header **and in the included cpp files**. Extreme cases like function calls in unrolled loops are not inlined, but loop unrolling has to be explicitly enforced via -funroll-loops.  

**Therefore adding inline to the function definition, as in this patch, does not change whether the function will be inlined or not**. But it helps the linker coping with multiple definitions.

I've tested the executable size on an ARMv9 architecture and I didn't see any difference with or without the inline. _This patch should therefore affect no user in terms of code size._

More info on this topic is e.g.:
http://stackoverflow.com/questions/3342816/code-in-header-files-will-always-be-inlined
http://stackoverflow.com/questions/5057021/why-are-c-inline-functions-in-the-header (the answer here clearly describes our situation).
The gcc manual states that

> [...] the effect on code size is less predictable; object code may be larger or smaller with function inlining, depending on the particular case.

http://gcc.gnu.org/onlinedocs/gcc/Inline.html

PR #77, opened 2013-10-02
Arduino Leonardo USB Serial Port

dawonn:
Added a Preprocessor option to use the USB serial port built into the Leonardo processor rather than the secondary UART. 

I think this should be the default but I don't want to break anyone's system. :)


dawonn:
Note that this is my first time using an Arduino, so forgive me if there is a better way to do this. :)

dawonn:
Fixed an issue, wasn't sure how to update this pull request so I just made another: 
https://github.com/ros-drivers/rosserial/pull/78

PR #78, opened 2013-10-02
Leonardo: Use the USB serial port for ROS messages option

dawonn:
Added the ability to use the USB serial port for ROS messages.

Note that I added a fail-safe start-up delay to prevent a condition where the serial link becomes flooded and the user wouldn't be able to upload new code without a hardware reset procedure that took me a few hours to figure out.... :)


bit-pirate:
Thanks for this PR @dawonn!

Please close #91, once this PR has been merged.

mikeferguson:
I did a cherry-pick on the to avoid pulling in this duplicate merge, see https://github.com/ros-drivers/rosserial/commit/888a7e5aa756817c877a9fd1fddba32dbc26337c

@PaulBouchier I'll let you do release.

Issue #8, opened 2013-02-07
rosserial_python: make more robust


mikeferguson:
- http://kforge.ros.org/rosserial/trac/ticket/17
- http://kforge.ros.org/rosserial/trac/ticket/55
- http://kforge.ros.org/rosserial/trac/ticket/73

mikeferguson:
17 and 73 are applied. 55:

 ROS: Electric
OS: OSX 10.7.3

No matter what example (e.g., hello world sketch) I run, it always crashes. I applied the patch to make rosserial compatible with Arduino 1.0.

[kevin@TARDIS rosserial]$ rosrun rosserial_python serial_node.py /dev/cu.usbserial-A4001lNd
[INFO] [WallTime: 1330743435.562765] ROS Serial Python Node
[INFO] [WallTime: 1330743435.568529] Connected on /dev/cu.usbserial-A4001lNd at 57600 baud
[INFO] [WallTime: 1330743438.066504] Note: publish buffer size is 280 bytes
[INFO] [WallTime: 1330743438.066825] Setup publisher on chatter [std_msgs/String]
[INFO] [WallTime: 1330743444.056641] Packet Failed : Failed to read msg data
[INFO] [WallTime: 1330743447.067020] Packet Failed : Failed to read msg data
[INFO] [WallTime: 1330743449.061603] Packet Failed : Failed to read msg data
[INFO] [WallTime: 1330743456.067540] Packet Failed : Failed to read msg data
[INFO] [WallTime: 1330743458.067177] Failed Packet Flags
[INFO] [WallTime: 1330743458.071638] Failed Packet Flags
[INFO] [WallTime: 1330743463.073970] Packet Failed : Failed to read msg data
[INFO] [WallTime: 1330743470.079938] Packet Failed : Failed to read msg data
Traceback (most recent call last):

```
File "/Users/kevin/ros_sandbox/rosserial/rosserial_python/nodes/serial_node.py", line 58, in <module>

    client.run()

File "/Users/kevin/ros_sandbox/rosserial/rosserial_python/src/rosserial_python/SerialClient.py", line 280, in run

    checksum = sum(map(ord,header) ) + sum(map(ord, msg)) + ord(chk)
```

TypeError?: ord() expected a character, but string of length 0 found

PR #80, opened 2013-10-10
fixes https://github.com/ros-drivers/rosserial/issues/8

xaedes:
when transmission errors caused the previous read request to timeout, serial_node crashed


mikepurvis:
LGTM.

xaedes:
Would be nice, if this could be accepted as it is a somehow serious issue as it crashes the node.

mikepurvis:
Yes, I agree—I don't have the commit privileges to merge it, though. @PaulBouchier ?

mikeferguson:
LGTM and appears to be quite similar to the other checks already implemented.

PR #88, opened 2013-11-26
Use stream logging in rosserial_server

mikepurvis:
to avoid problems with printfing size_t and similar types.


mikeferguson:
lgtm

PR #90, opened 2013-12-16
rosserial_arduino support for Teensy 3.0 and 3.1

kjanesch:
Added an extra #if to ArduinoHardware.h to check for Teensy 3.x compiler flags (**MK20DX128/256**) when using Teensyduino. Ensures that the USB serial connection is used instead of an onboard UART.

Tested and works with Arduino 1.0.5 and Teensyduino 1.15-1.17 on Ubuntu 13.10 with ROS Hydro.


mikeferguson:
LGTM

PR #99, opened 2014-02-18
add better debugging information when packages are missing dependencies

mikeferguson:
Patch generated while figuring out http://answers.ros.org/question/129701/rosserial-make_libraries-missing-custom-message/


mikepurvis:
LGTM

PR #102, opened 2014-03-02
fix two points: 1. the number of bytes to read for chk_byte, 2. the wron...

tongtybj:
...g indentation about the defination of sendDiagnostics()


tongtybj:
Fix two bugs related to python rosserial server.
1. The number of bytes while reading the chk_byte 
2.  The indentation error around the definition of  sendDiagnostics()

PR #103, opened 2014-03-03
Support ID_TX_STOP

mikepurvis:
Merging this fixes #57, from @tongtybj.

I've done some tests with a Leonardo; it seems to work well. @mikeferguson, @PaulBouchier, are you guys fine with this going into hydro-devel? There should be perfect interoperability between Hydro firmwares before and after this change and serial_node before and after.


mikeferguson:
Have you confirmed that indeed the message checksum does not change with the added comment? (It shouldn't, but I haven't checked that there are no regressions in either our client library or the rospy one)

rosserial_python/src/rosserial_python/SerialClient.py
```diff
@@ -471,10 +501,18 @@ def setupSubscriber(self, data):
...
-            self.subscribers[msg.topic_name] = sub
-            self.setSubscribeSize(msg.buffer_size)
-            rospy.loginfo("Setup subscriber on %s [%s]" % (msg.topic_name, msg.message_type) )
+            if not msg.topic_name in self.subscribers.keys():
+                sub = Subscriber(msg, self) 
+                self.subscribers[msg.topic_name] = sub
+                self.setSubscribeSize(msg.buffer_size)
+                rospy.loginfo("Setup subscriber on %s [%s]" % (msg.topic_name, msg.message_type) )
+            elif msg.message_type != self.subscribers[msg.topic_name].message._type:
+                    old_message_type = self.subscribers[msg.topic_name].message._type
```
mikeferguson: Indentation seems wrong on either this block, or the part in the "if" statement above.

mikepurvis:
Yes, it's the same: https://gist.github.com/mikepurvis/9334306

The 63aa5e8f1bdd6f35c69fe1a1b9d28e9f value given is the same as what you see with `rosmsg show rosserial_msgs/TopicInfo` on the currently-released rosserial_msgs.

rosserial_python/src/rosserial_python/SerialClient.py
mikepurvis: Good catch, thanks—fixed now.

(though it would also have been caught by that roslint patch that will go in after this!)

mikeferguson:
lgtm

mikepurvis:
When the indigo-devel branch is created, rosserial_msgs can be modified to uncomment the enum for ID_TX_STOP.

Issue #105, opened 2014-03-12
wrong message lenght, if message size is more than 255 bytes


linvinus:
wrong message lenght, if message size is more than 255 bytes
http://answers.ros.org/question/73807/rosserial-deserialization-error/

fixed in https://github.com/linvinus/rosserial/commit/b5219aad08fa844b8e5cd0c4708c2acb0fb8634f

mikepurvis:
Closing in favour of #106

PR #106, opened 2014-03-12
fix: wrong message lenght, if message size more than 255

linvinus:
fix: #105
fix odometry deserialization error http://answers.ros.org/question/73807/rosserial-deserialization-error/

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -419,13 +419,13 @@ namespace ros {
 	  return 0;
 
         /* serialize message */
-        int l = msg->serialize(message_out+7);
+        unsigned int l = msg->serialize(message_out+7);
 
         /* setup the header */
         message_out[0] = 0xff;
         message_out[1] = PROTOCOL_VER;
-        message_out[2] = (unsigned char) l&255;
-        message_out[3] = (unsigned char) l>>8;
+        message_out[2] = (unsigned char) ((unsigned int)l&255);
+        message_out[3] = (unsigned char) ((unsigned int)l>>8);
 	message_out[4] = 255 - ((message_out[2] + message_out[3])%256);
         message_out[5] = (unsigned char) id&255;
         message_out[6] = ((unsigned char) id>>8);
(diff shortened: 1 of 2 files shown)


mikepurvis:
This looks great, thanks for the fix. Will test and merge later today.

rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -419,16 +419,16 @@
...
         /* serialize message */
-        int l = msg->serialize(message_out+7);
+        unsigned int l = msg->serialize(message_out+7);
 
         /* setup the header */
         message_out[0] = 0xff;
         message_out[1] = PROTOCOL_VER;
-        message_out[2] = (unsigned char) l&255;
-        message_out[3] = (unsigned char) l>>8;
+        message_out[2] = (unsigned char) ((unsigned int)l&255);
```
mikeferguson, line 427: We can probably remove the (unsigned int) here, since l is already an unsigned, no? Same on the next line.

mikeferguson:
Generally lgtm.

@linvinus in the future, if you have a Pull Request ready, just fire that off, no need to spend extra time opening a bug report and then immediately opening a pull request to fix.

linvinus:
@mikeferguson of course, i'm sorry.

rosserial_client/src/ros_lib/ros/node_handle.h
linvinus, line 427: yes, we can.
this version also will be correct.

unsigned int l = msg->serialize(message_out+7);

/\* setup the header */
message_out[0] = 0xff;
message_out[1] = PROTOCOL_VER;
message_out[2] = (unsigned char) (l&255);
message_out[3] = (unsigned char) (l>>8);
message_out[4] = 255 - ((message_out[2] + message_out[3])%256);
message_out[5] = (unsigned char) (id&255);
message_out[6] = (unsigned char) (id>>8);

mikepurvis:
@mikeferguson Strong opinions about this going into hydro as well vs. just indigo?

mikeferguson:
It's actually a bug fix -- so it should go into both hydro & indigo.

PR #107, opened 2014-03-27
add install target for make_library.py

paulbovbel:


paulbovbel:
Found associated thread http://answers.ros.org/question/119031/rosserial_client-does-not-include-make_librarypy/, will fix tutorial.

mikeferguson:
@mikepurvis looks like your comment is gone, but I just wanted to point out that rosserial_client/make_library.py should not be executed, it is used by the various rosserial_PLATFORM/make_libraries.py executables. This was part of the catkin refactor (the old approach no longer worked well with catkin, so I ended up making rosserial work closer to a message generator package, and at the same time made it easier for new platforms to be created).

mikepurvis:
Yes, sorry— I saw Paul's update after my reply and removed it. Makes perfect sense.

PR #108, opened 2014-04-10
make tf topic absolute instead of relative

paulbovbel:
prevents accidental remapping via <group> tag in roslaunch


mikeferguson:
Seems reasonable. +1

PR #114, opened 2014-04-29
Use const char* instead of char* for strings in messages.

mikepurvis:
The present implementation has GCC complaining thusly:

```
warning: deprecated conversion from string constant to 'char*' [-Wwrite-strings]
     feedback_msg_.header.frame_id = "base_link";
```

Seems to compile and do the right thing for me.


mikepurvis:
@PaulBouchier, @mikeferguson Do you foresee any problems with this change?

PR #115, opened 2014-04-29
No xx_val pointers for fixed-length arrays of messages.

mikepurvis:
Eliminate these unused _val pointers—they're local-scope vars which do nothing as far as I can tell (apart from throw up a compiler warning about an unused local variable).


mikepurvis:
@PaulBouchier Any objections here?

PR #117, opened 2014-05-09
Remove include of ros.h from time.cpp

mikepurvis:
It compiles fine without it— it being here is a pain because there's an assumption that the client library's "ros.h" file will be in the include path, whereas a client library might prefer it as "ros/ros.h" or something else entirely.

Is there some historical reason this was here? @mikeferguson


mikeferguson:
Its inclusion would appear to be a mistake -- this LGTM

PR #123, opened 2014-05-21
rosserial_windows - package for building rosserial clients on Windows

kshehata:
I've added the rosserial_windows package for building rosserial_clients for Windows using Visual Studios. There's a tutorial on how to use the library on the ROS wiki http://wiki.ros.org/rosserial_windows/Tutorials/Hello%20World#preview

This is based on Hydro because my current environment is Hydro, but should be an easy port if not identical on Indigo.


rosserial_windows/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(rosserial_windows)
+
+find_package(catkin REQUIRED)
+catkin_package(CATKIN_DEPENDS)
+
+install(DIRECTORY src/ros_lib
+  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/src
```
mikepurvis, line 8: This seems a bit hinky. Does it actually work from the installed version? ie, 

```
catkin_make install
source install/setup.bash
rosrun ...
```

rosserial_windows/src/examples/MessageReceive/messagereceive.cpp
```diff
@@ -0,0 +1,35 @@
+#include "stdafx.h"
+#include <string>
+#include <stdio.h>
+#include "ros.h"
+#include <geometry_msgs/PoseWithCovarianceStamped.h>
+#include <windows.h>
```
mikepurvis, line 31: What's the scheme here for quotes vs. angle-brackets? The style guide is silent on this type of thing, but please at least be consistent.

rosserial_windows/src/examples/MessageReceive/messagereceive.cpp
```diff
@@ -0,0 +1,35 @@
...
+using std::string;
+
+void estimated_pose_callback(const geometry_msgs::PoseWithCovarianceStamped& pose) {
+        printf("Received pose %f, %f, %f\n", pose.pose.pose.position.x,
+                pose.pose.pose.position.y, pose.pose.pose.position.z);
+}
+
+int _tmain(int argc, _TCHAR* argv[])
+{
+        ros::NodeHandle nh;
```
mikepurvis: 2 spaces instead of tabs, please.

rosserial_windows/src/examples/MessageReceive/messagereceive.cpp
```diff
@@ -0,0 +1,35 @@
+#include "stdafx.h"
```
mikepurvis, line 26: I know it's just an example, but there should still be a BSD title block on this file.

mikepurvis:
Made a couple of notes for stylistic fixes, but overall this is great— thanks, @kshehata.

Any concerns about this, @mikeferguson, @PaulBouchier ?

rosserial_windows/package.xml
```diff
@@ -0,0 +1,27 @@
+<package>
+  <name>rosserial_windows</name>
+  <version>0.0.0</version>
```
mikeferguson, line 3: @mikepurvis this may need to be made 0.5.5 for the next release to work correctly?

rosserial_windows/CMakeLists.txt
mikeferguson, line 8: This isn't source you would run with rosrun, this exactly line is used in all the rosserial libraries in order to copy over the source that is needed for the arduinos -- the make_library script then accesses that source.

rosserial_windows/src/ros_lib/WindowsSocket.cpp
```diff
@@ -0,0 +1,233 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2014, Clearpath Robotics Inc.
```
mikeferguson: Is Clearpath right here?

rosserial_windows/src/ros_lib/WindowsSocket.cpp
```diff
@@ -0,0 +1,233 @@
...
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *  * Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  * Redistributions in binary form must reproduce the above
+ *    copyright notice, this list of conditions and the following
+ *    disclaimer in the documentation and/or other materials provided
+ *    with the distribution.
+ *  * Neither the name of Willow Garage, Inc. nor the names of its
```
mikeferguson: Willow here? yeah, check the attribution on all these license headers.

mikeferguson:
Other than licensing and style, LGTM. Gotta say, this was never a use case I imagined when creating rosserial... cool idea.

rosserial_windows/src/examples/MessageReceive/messagereceive.cpp
kshehata, line 31: I've gone with quotes for anything that's in the project itself, and angle brackets for anything that is in include directories. Though ros.h would at first appear to be a system library is actually in the project, hence quotes.

rosserial_windows/src/examples/MessageReceive/messagereceive.cpp
kshehata: Visual studio defaults to tabs. I was going for what would be most commonly used in Windows. I can change it to spaces, but anyone who goes to use in VS will likely fight with their editor.

rosserial_windows/src/ros_lib/WindowsSocket.cpp
kshehata: I basically rewrote the WindowsSocket code. Hence the Clearpath on this file.

rosserial_windows/src/ros_lib/WindowsSocket.cpp
kshehata: Gah. I copied and pasted license text, though I did re-write this file. I'll correct the license header.

rosserial_windows/src/examples/MessageReceive/messagereceive.cpp
mikepurvis: Does it? Not the impression I get from this article: http://lostechies.com/jimmybogard/2012/03/07/tabs-versus-spaces-spaces-won/

It's good to cooperate with the local convention, but ROS style is still 2 spaces. A user building on rosserial in Windows is still creating ROS ecosystem code, so it really should be according to ROS conventions. Especially because a rosserial_windows user might not even necessarily be using MSVC— they could be on CodeBlocks, Eclipse, etc.

Whichever way it goes, a note on the wiki about it is warranted.

rosserial_windows/package.xml
mikepurvis, line 3: Yes it will. Might make sense to change it post merge as part of the release preparation, not sure.

rosserial_windows/src/ros_lib/WindowsSocket.cpp
mikepurvis: Haha, whoops. @kshehata, we have a standard Clearpath BSD title block [here](https://github.com/clearpathrobotics/headers/blob/master/bsd/cpp).

rosserial_windows/src/ros_lib/WindowsSocket.cpp
mikepurvis: Not sure—is there a standard about this? At present, rosserial_server is copyright Clearpath. A lot of stuff is copyright Willow, OSRF, individual contributors. It gets a bit murky with multiple packages in a single repo. @tfoote, opinions on how this type of thing should be handled?

kshehata:
Yea, it's a pretty cool use of the tool @mikeferguson. Have to admin: I didn't actually come up with this myself. Some of the other guys here at Clearpath came up with it, and I just cleaned it up from a hack to distributable code.

rosserial_windows/src/examples/MessageReceive/messagereceive.cpp
kshehata: Good point. I've changed the style to match ROS guidelines using indent. There are plenty of examples on the web for how to change the defaults in VS to be not-stupid.

kshehata:
Just pushed the changes to the licenses and the style for the examples.

mikeferguson:
Also, it might make sense to squash these commits into one -- I presume intermediate versions do not work.

kshehata:
Most of the intermediate versions should work, though if it's easier on your end, I can certainly squash it into one big commit.

kshehata:
Just following up on this: do you want this redone as one big commit, or is the current format ok?

PR #126, opened 2014-06-09
Added the missing inWaiting() to RosSerialServer

girtslinde:
"serial_node.py tcp" was crashing due to the missing inWaiting(), which an actual serial port would have.


mikepurvis:
LGTM

PR #129, opened 2014-08-05
Service client support for rosserial-server

jjekircp:
Allows rosserial clients to register service clients and make service calls. Only tested with rosserial-windows so far.

### DIFF
--- a/rosserial_msgs/CMakeLists.txt
+++ b/rosserial_msgs/CMakeLists.txt
@@ -11,9 +11,8 @@ add_message_files(FILES
 add_service_files(FILES
                   RequestParam.srv
                   RequestMessageInfo.srv
+                  RequestServiceInfo.srv
                  )
 
 generate_messages()
-
 catkin_package(CATKIN_DEPENDS message_runtime)
-
--- a/rosserial_msgs/srv/RequestServiceInfo.srv
+++ b/rosserial_msgs/srv/RequestServiceInfo.srv
@@ -1,0 +1,6 @@
+# service name
+string service
+---
+# If found, return md5 string of system's version of the service, and
+# textual definition. If not found, both strings are to be empty.
+string md5
--- a/rosserial_msgs/srv/RequestServiceInfo.srv
+++ b/rosserial_msgs/srv/RequestServiceInfo.srv
@@ -1,6 +1,4 @@
 # service name
 string service
 ---
-# If found, return md5 string of system's version of the service, and
-# textual definition. If not found, both strings are to be empty.
 string md5
--- a/rosserial_msgs/srv/RequestServiceInfo.srv
+++ b/rosserial_msgs/srv/RequestServiceInfo.srv
@@ -1,4 +1,6 @@
 # service name
 string service
 ---
-string md5
+string service_md5
+string request_md5
+string response_md5
--- a/rosserial_python/nodes/message_info_service.py
+++ b/rosserial_python/nodes/message_info_service.py
@@ -70,15 +70,14 @@ class MessageInfoService(object):
 
   def _service_info_cb(self, req):
     rospy.logdebug("req.service is %s" % req.service)
-    package_message = tuple(req.service.split("/"))
-    if not self.service_cache.has_key(package_message):
-      rospy.loginfo("Loading module to return info on service %s/%s." % package_message)
-      srv = load_service(*package_message)
-      self.service_cache[package_message] = (srv._md5sum)
+    package_service = tuple(req.service.split("/"))
+    if not self.service_cache.has_key(package_service):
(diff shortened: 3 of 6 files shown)


mikepurvis:
On first glance, this looks really good. Please describe, at a high level, how this implementation handles various situations such as:
- Client calls service multiple times without waiting for responses inbetween.
- Client calls service and the response has a bit error and falls on the floor. 
- Multiple clients are connected, calling the same service at once.
- The TopicInfo message setting up the service client falls on the floor. Would be great to see service client support added to the rosserial_server's [Require API](http://wiki.ros.org/rosserial_server#Require_API).

jjekircp:
Client calls service multiple times without waiting for responses inbetween.
- My callback (when I receive the message from the client asking me to
  perform the service call) blocks on the call into roscpp that actually
  performs the service call
- Every time I receive a service request I send a service response
  (though it may be a blank ShapeShifter message if the service call failed).
- I'm assuming that incoming service calls are queued up in the network
  stack and that my callback will get called over and over again as service
  calls come in. I expect that that case should work fine.

Client calls service and the response has a bit error and falls on the
floor.
- If this problem happens when I call into roscpp? I send the client
  whatever roscpp put in its response message. This may end up being a blank
  ShapeShifter message.
- If this problem happens between rosserial_server and the
  rosserial_client? The client will continue waiting for the response, and it
  may (pretty sure rosserial_windows does this) wait forever.

Multiple clients are connected, calling the same service at once.
- I'm not sure I fully understand how multiple clients are handled. It
  looks like they each get their own Session instance, which contains all of
  the publisher, subscriber, and service client instances. It also looks to
  me like rosserial_server is completely single-threaded asynchronous.
  Consequently, I expect Boost's async IO library to call the callback for
  one of the incoming TCP messages before the other (i.e. one client before
  the other), and whichever comes first will make the service call in roscpp
  (see earlier answer), send the response to its client, then the other
  client service call will go through.

The TopicInfo message setting up the service client falls on the floor.
Would be great to see service client support added to the
rosserial_server's Require API
http://wiki.ros.org/rosserial_server#Require_API.
- Setting up a service client takes two TopicInfo messages: a service
  client publisher message (so the client can publish the request) and a
  service client subscriber message (so the client can subscribe to the
  response). If the service client subscriber message goes through but the
  service client publisher message falls on the floor, the service call
  registration as well as the service call itself will both succeed, with the
  caveat that the service request message MD5 will not be verified. If the
  service client subscriber message falls on the floor but the service client
  publisher message goes through, the service call registration will appear
  to succeed but a subsequent service call will not be successful.
- Specifically, when the rosserial client creates its service client
  object, it registers a callback for the service response message and
  assigns it a TopicID. This TopicID is used in the service client subscribe
  message, and rosserial_server saves it and uses it when sending the service
  call response message to the rosserial client. If this TopicID is not sent,
  the rosserial client will not recognize and then discard the service call
  response message, and continue waiting (probably indefinitely) for the
  response message that it will never see.

On Tue, Aug 5, 2014 at 1:00 PM, Mike Purvis notifications@github.com
wrote:

> On first glance, this looks really good. Please describe, at a high level,
> how this implementation handles various situations such as:
> - Client calls service multiple times without waiting for responses
>   inbetween.
> - Client calls service and the response has a bit error and falls on
>   the floor.
> - Multiple clients are connected, calling the same service at once.
> - The TopicInfo message setting up the service client falls on the
>   floor. Would be great to see service client support added to the
>   rosserial_server's Require API
>   http://wiki.ros.org/rosserial_server#Require_API.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/ros-drivers/rosserial/pull/129#issuecomment-51227167.

## 

_Jonathan Jekir_ | Software Engineer | Clearpath Robotics, Inc.
1425 Strasburg Rd., Unit 2A, Kitchener, Ontario N2C 1L3 Canada
_T: \* 519-513-2416 x854 | *E:_ jjekir@clearpathrobotics.com
email@clearpathrobotics.com  | _W:_ www.clearpathrobotics.com

*Your Unmanned Experts *
*TM *

mikepurvis:
Okay, going to merge this with the understanding that the support is non-robust for an unreliable (serial) transport, but further acknowledging that much of the burden for robustifying it falls on the client, not the server.

mikepurvis:
Merged to `indigo-devel` via af7749bba16a9c476d26fc304d1bfa53eec550cc.

jjekircp:
Sounds good. This is in the indigo-devel branch now...how can I use it with ROS hydro?

mikepurvis:
Add the indigo-devel branch to your source workspace; ideally, this would be managed using [wstool](http://wiki.ros.org/wstool).

PR #131, opened 2014-08-09
fix-headers

mikepurvis:
This is for compliance with ROS Style, but also assists with making rosserial_server embeddable.


mikeferguson:
LGTM

Issue #128, opened 2014-08-04
rosserial_client does not handle a disconnection from the server elgantly


jscatena88:
When the server goes down (e.g for a restart etc) the rosserial client will just get stuck in the spinOnce() state machine's while(true) loop with mode_ = MODE_FIRST_FF (i.e. 0). For the use case of a potentially remote system it would be nice to be able to respond gracefully to the server coming up and down while the remote machine stays on. 

I fixed it in our code base by changing(@ line 209 in node_handle.h):

```
    }else if( mode_ == MODE_FIRST_FF ){
        if(data == 0xff){
          mode_++;
          last_msg_timeout_time = c_time + MSG_TIMEOUT;
        }
```

to: 

```
    }else if( mode_ == MODE_FIRST_FF ){
        if(data == 0xff){
          mode_++;
          last_msg_timeout_time = c_time + MSG_TIMEOUT;
        }
        else if (difftime(hardware_.time(), c_time) > (SYNC_SECONDS)){
          //Been here too long, return -1
          return -1;
    }
```

With that code in place my code looks for the -1 to be returned by spinOnce() and exits if it is seen.

If this fix seems generally applicable feel free to add it, or I can commit it.

jscatena88:
I actually have it return a -2 now so I can differentiate it from another return -1 in the statemachine

mikepurvis:
I bring my server up and down all the time and have not been bitten by this, but it looks like a good change. Can you create a pull request against `indigo-devel` for review?

jscatena88:
Sorry I haven't checked git hub in a few days. I should be able to do that this weekend

jscatena88:
Just made the pull request. Barring any problems with the code I will close this issue once it is merged

PR #132, opened 2014-08-11
added time out to the state machine

jscatena88:
Implmented as indicated in issue #128

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -211,6 +211,9 @@ namespace ros {
               mode_++;
               last_msg_timeout_time = c_time + MSG_TIMEOUT;
             }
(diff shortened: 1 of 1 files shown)


mikepurvis:
Should we also set `configured_ = false`?

Also, that line is now really long. I know it's consistent with the surrounding conditionals, but could you please move the comment to right above the return, inside the braces?

jscatena88:
Yeah setting configured_ false seems like a good idea as well.

And no problem with the comment, I wasn't sure wether to prioritize consistency or reasonable line lengths.

mikepurvis:
Looks great, thanks!

mikepurvis:
Ugh, this is causing a problem for me now due to difftime. Is there any reason to use that vs. just a straight subtraction? The `time()` function should either return an unsigned integer value or something which supplies `operator-`.

Thoughts?

PR #136, opened 2014-08-21
Remove difftime call from rosserial_client

mikepurvis:
Fixes a regression from #132.

@jscatena88, does this work okay for you?


jscatena88:
Yes that works fine for me, I only used difftime because I am primarily a linux/unix developer.

PR #138, opened 2014-09-10
CMake Helpers

mikepurvis:
This PR resurrects the old rosbuild rosserial functionality of being able to build a firmware right inside your ROS package. Example package you can build with this change:

https://github.com/mikepurvis/hello_arduino

Some basic wiki documentation will be created for the three CMake functions, as well as filling out the catkin side of [this tutorial](http://wiki.ros.org/rosserial_arduino/Tutorials/CMake).


rosserial_client/scripts/make_libraries
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python
+
+#####################################################################
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2014, Mike Purvis
```
mikeferguson, line 7: This file looks like a slightly modified version of https://github.com/ros-drivers/rosserial/blob/indigo-devel/rosserial_arduino/src/rosserial_arduino/make_libraries.py -- right? It should retain the 2013 Willow copyright (you can certainly add your 2014 copyright)

mikepurvis, line 7: Excellent point, thanks. Fixed now.

mikepurvis:
Would like to merge this in the next few hours— I don't think there's much that should break anything here. Any further thoughts/concerns?

mikepurvis:
Tutorial updated: http://wiki.ros.org/rosserial_arduino/Tutorials/CMake

mikeferguson:
Maybe I'm just missing it, but it looks like you are using a submodule for the arduino cmake stuff -- does this get initialized automatically somehow during the build process? If not, this only works from source?

mikepurvis:
Huh, I was certain that bloom now supported submodules, however I see that the relevant ticket is still open: https://github.com/ros-infrastructure/bloom/issues/217

So yes, if this doesn't work then either the submodule could be initialized as an execute_process during the configure step, or the files could just be pasted in. I hate maintaining forks, but maybe that would be best in this case. Thoughts?

mikepurvis:
Welp, I tried it out and indeed, bloom does not capture submodule content as part of the upstream— not sure how I got the impression that was the case.

I've added a new commit to remove the submodule and just commit the files straight-up.

mikeferguson:
LGTM

Issue #86, opened 2013-11-14
Messages in geometry_msgs not handling negative numbers


avardy:
I tested geometry_msgs/Twist and geometry_msgs/Vector3 and there seems to be some error in the serialization which turns all negative numbers (I tested -100, -10) into -0.

As a hack to control my robot (iRobot Create with custom Arduino-based control board) I am sending Point32 messages to encode forward speed and rotation.

mikeferguson:
@PaulBouchier if you end up looking into this, it is likely a regression in the Float64 support from recent updates. The Float64 handler is quite custom on the Arduino since there is no 64-bit types so we end up converting to 32-bit ones. We used this script as a test: https://github.com/ros-drivers/rosserial/blob/hydro-devel/rosserial_arduino/src/ros_lib/tests/float64_test/float64_test.pde

PaulBouchier:
Thanks for the pointer Michael. I'm trying to get the arduino Due won't compile bug fixed, but I'll look at this after.

PR #139, opened 2014-09-18
Fixes Issue #86

mgerdzhev:
Fixed the deserialize() of AVR_Float64DataType in order to support negative numbers.
This should resolve issue #86


mikepurvis:
Hi @mgerdzhev, thanks for the contribution! Can you explain what your process was for testing and verifying this?

mgerdzhev:
hi @mikepurvis. I am using the Float64DataType as part of geometry_msgs/Twist. If I send it negative numbers and print them out they always show up as -0.0 due to the sign bit not properly being set. With the current change, the sign bit is copied from the 64bit to the 32bit representation and my Twist messages are received and printed properly (negative and positive). I tested it out on our robot and the message is interpreted properly now.

mikeferguson:
This seems reasonable, although I'm entirely sure how this came to be broken in the first place (it did used to work)...

@mikepurvis you can do a quick test locally with this sketch: https://github.com/ros-drivers/rosserial/blob/hydro-devel/rosserial_arduino/src/ros_lib/tests/float64_test/float64_test.pde which lets you send float messages down, subtract 1 and return it.

mikepurvis:
Okay, should be able to get to this on the weekend, thanks again.

mikepurvis:
This fix is merging as part of #144.

PR #144, opened 2014-09-27
Move AVR serialization logic, add unit test

mikepurvis:
This PR includes and supercedes #139.

As an additional obvious benefit, the serialization logic for Float64, if used, will appear only once in the binary, rather than once per field using it. If unused, it will be optimized away.

@mikeferguson, @PaulBouchier, @mgerdzhev, this look okay to you guys?

### DIFF
--- a/rosserial_client/CMakeLists.txt
+++ b/rosserial_client/CMakeLists.txt
@@ -14,3 +14,8 @@ install(DIRECTORY src/ros_lib
 install(PROGRAMS scripts/make_libraries
   DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
 )
+
+if(CATKIN_ENABLE_TESTING)
+  include_directories(src)
+  catkin_add_gtest(float64_test test/float64_test.cpp)
+endif()
--- a/rosserial_client/src/ros_lib/ros/msg.h
+++ b/rosserial_client/src/ros_lib/ros/msg.h
@@ -35,18 +35,95 @@
 #ifndef _ROS_MSG_H_
 #define _ROS_MSG_H_
 
+#include <stdint.h>
+
 namespace ros {
 
-  /* Base Message Type */
-  class Msg
-  {
-    public:
-      virtual int serialize(unsigned char *outbuffer) const = 0;
-	  virtual int deserialize(unsigned char *data) = 0;
-      virtual const char * getType() = 0;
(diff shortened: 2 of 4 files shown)


mikeferguson:
LGTM

mikepurvis:
I actually _still_ haven't had a chance to do a quick test of this on an actual Arduino— that'll happen before a merge.

mgerdzhev:
From the changes it still looks like you need to add my changes in lines 106-109 of the msg.h file.

mikepurvis:
I'm sorry about that! I definitely merged it first, but perhaps failed to regenerate the message header that formed the basis of the `msg.h` implementation.

The test intentionally include negative numbers— why is it passing without the fix?

mikeferguson:
@mgerdzhev This code handles the negative numbers, just slightly different from your formulation or the original rosserial code, and I think the way @mikepurvis has done it makes it more readable than either of the previous versions as well.

mgerdzhev:
@mikeferguson, @mikepurvis I just tested it on my arduino and negative numbers are not handled properly again with this change. 
Adding this line instead of the if statement in lines 117-120 of the msg.h file fixes the problem. 
 *val |= ((uint32_t)(*(inbuffer++)) & 0x80)<<24;

mikeferguson:
@mikepurvis DId you run this on hardware? If so, what compiler version? @mgerdzhev what compiler version of avr-gcc are you on?

mikepurvis:
I did a similar test this morning and found the same thing. Thanks for the fix— I've confirmed this is good.

I'm still baffled as to why the PC test doesn't catch the problem; it may be that the previous version relied on undefined behaviour, and the PC's more robust float logic could handle it. In any case, fix is done, will push and merge shortly.

(I'm on avr-gcc 4.8.2)

mgerdzhev:
I'm using gcc-avr-1:4.8-2.1

mikepurvis:
Okay, rebased in preparation for merge.

gavanderhoorn:
@mikepurvis: I'm not too familiar with the release schedule of `rosserial`: has the fix for Float64 been backported to Hydro?

mikepurvis:
No, I haven't been backporting anything. I would accept a non-breaking PR against `hydro-devel`, but I don't have the bandwidth to do it myself.

gavanderhoorn:
Ok, understood. Thanks.

PR #146, opened 2014-10-24
Updated size checksum verification to properly incorporate the high byte...

jjekircp:
... of the message size for messages larger than 0x00FF (255) characters


mikepurvis:
To clarify, this is an issue for a message with a length > 255 sent from the server and received by the client?

rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -228,13 +228,14 @@ namespace ros {
             bytes_ = data;
             index_ = 0;
             mode_++;
-            checksum_ = data;               /* first byte for calculating size checksum */
+            size_checksum_ = data;               /* first byte for calculating size checksum */
           }else if( mode_ == MODE_SIZE_H ){   /* top half of message size */
             bytes_ += data<<8;
+            size_checksum_ |= data;
```
mikeferguson, line 234: This should be +=, no?

jjekircp:
That would make more sense, but I wanted to match rosserial_server which has an |= and doesn't seem to work if I put +=.

mikepurvis:
Interesting, that sounds like a bug in rosserial_server, then. In the absence of canonical documentation about how the checksum is supposed to work, I would treat the `rosserial_python` implementation as the reference, which uses a sum:

https://github.com/ros-drivers/rosserial/blob/indigo-devel/rosserial_python/src/rosserial_python/SerialClient.py#L687

Include a patch to rosserial_server in this PR?

mikeferguson:
The checksum was based on the Dynamixel checksum, thus, it should be +=

jjekircp:
I've amended this pull request to include the change to rosserial_server.

@mikepurvis In response to your earlier question: Yes, this is an issue with a message of length >255 sent from the server and received by the client.

mikepurvis:
But the client should also use a + in this scenario. Can you describe how you verified that this is doing the right thing?

mikepurvis:
@jjekircp Let's get this merged. Can we get the client and server in sync and ready to go in here?

mikepurvis:
Resolved in #173.

PR #149, opened 2014-11-21
Support OS X time in the embeddedlinux port.

mikepurvis:
@PaulBouchier, any concerns on this?


mikepurvis:
Manually merged into `jade-devel`.

PR #151, opened 2014-12-01
Use native 64-bit double on embeddedlinux.

mikepurvis:
@PaulBouchier Is this likely to cause problems for any of your platforms?

My impression is that most OpenWRT-type are based on 32-bit chips like AR7130, and wouldn't need the float translation logic necessary for 8-bit platforms.


PaulBouchier:
Seems like a safe bet Mike. double is universal on Linux AFIK.
Paul

PR #152, opened 2014-12-01
Generate constructors for messages.

mikepurvis:
This is going to bloat the microcontroller firmwares a wee bit, but right now we depend on undefined behaviour in that we assume that messages are going to be allocated on zeroed memory. It's particularly problematic for the length field on a subscriber to be non-zero, since that really confuses the logic that triggers a realloc when a larger sized array is received.

Situations where the message memory may not be zeroed: messages allocated on the stack, clients running on PCs, unit tests, etc.

### DIFF
--- a/rosserial_client/src/rosserial_client/make_library.py
+++ b/rosserial_client/src/rosserial_client/make_library.py
@@ -79,6 +79,9 @@ class PrimitiveDataType:
         self.type = ty
         self.bytes = bytes
 
+    def make_initializer(self, f, trailer):
+        f.write('      %s(0)%s\n' % (self.name, trailer))
+
     def make_declaration(self, f):
         f.write('      %s %s;\n' % (self.type, self.name) )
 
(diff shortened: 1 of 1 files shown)


rosserial_client/src/rosserial_client/make_library.py
```diff
@@ -193,6 +209,12 @@ def __init__(self, name, ty, bytes, cls, array_size=None):
         self.size = array_size
         self.cls = cls
 
+    def make_initializer(self, f, trailer):
+        if self.size == None:
+            f.write('      %s_length(0), %s(NULL)%s\n' % (self.name, self.name, trailer))
+        else:
+            f.write('      %s()%s\n' % (self.name, self.name, trailer))
```
mikepurvis, line 216: This is a bit of a cop-out, but doing it properly is a lot more complex— you'd have to figure out if the array is of primitives or not, and insert a memset call in the constructor body in the case of primitives.

rosserial_client/src/rosserial_client/make_library.py
```diff
@@ -401,7 +432,7 @@ def __init__(self, name, package, definition, md5req, md5res):
         self.name = name
         self.package = package
 
-        sep_line = None
+        sep_line = len(definition)
```
mikepurvis, line 435: This is actually a fix for the issue which prevented map_msgs from compiling— the problem was services which defined only the request portion and not response. For example: http://docs.ros.org/indigo/api/map_msgs/html/srv/ProjectedMapsInfo.html

This change allows the generator to assume that in the absence of the dividing `---`, the divider is at the end of the service definition.

PR #155, opened 2014-12-24
we don't need roslib.load_manifest any more in catkin system

eisoku9618:
When I use rosserial and custom service, "cannot import package" error occurs.
By this PR, I can use a custom service.

( Is this related to http://answers.ros.org/question/11023/rosserial-and-custom-message-cannot-load-package-error/ ?)

### DIFF
--- a/rosserial_python/src/rosserial_python/SerialClient.py
+++ b/rosserial_python/src/rosserial_python/SerialClient.py
@@ -37,6 +37,7 @@ __author__ = "mferguson@willowgarage.com (Michael Ferguson)"
 
 import roslib;
 import rospy
+import imp
 
 import thread
 import multiprocessing
@@ -56,14 +57,10 @@ import signal
 
 def load_pkg_module(package, directory):
     #check if its in the python path
-    in_path = False
     path = sys.path
-    pkg_src = package+'/src' #check for the source directory which
-                             # is added to path by roslib boostrapping
-    for entry in sys.path:
-        if pkg_src in entry:
-            in_path = True
-    if not in_path:
+    try:
+        imp.find_module(package)
+    except:
         roslib.load_manifest(package)
     try:
         m = __import__( package + '.' + directory )


k-okada:
+1

mikepurvis:
This seems... too easy. Is there a downside here?

Also, do you definitely need this in Hydro? I'm wondering if we can merge and release to Indigo only, and then if you can use the Indigo branch from source in Hydro if needed.

mikeferguson:
This should work with all catkin-ized stuff since we now do a regular filesystem and paths. We couldn't do this in the old days of rosbuild since the path was custom setup based on the manifest (so something outside the manifest would be outside the python path). I'm not 100% sure what happens with non-catkinized stuff, but I think this will work -- I would recommend testing it on a rosbuild package though, since I know there a bunch of rosbuild users still using rosserial.

k-okada:
we're using them in Hydro, since most of "real" robot does not support
indigo, for example, pr2 is just finished catinized and we're currently
"beta testing" pr2/hydro package. Real robots has mass and inertia...
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/pr2_users/bASxfa_tPvw/pOSlk8VPQt4J

## 

◉ Kei Okada

eisoku9618:
I found that this script before this PR works well in general catkin system.

The reason why I thought that this script do not work in hydro/catkin for the first time is that our laboratory packages contain both package.xml and manifest.xml for a reason. This was just a personal matter. (I'm sorry for the lack of confirmation.)

So,  I modify this PR to switch whether to use `load_manifest` or not depending on the build system.
By this PR, catkin users can skip unnecessary `load_manifest`.

mikepurvis:
Any objections to pulling this in for Jade?

mikepurvis:
Merged into `jade-devel`: 43ee6bd349606e4dcb4467e3634f9450c84c592e

Issue #161, opened 2015-02-11
Running rosrun rosserial_arduino make_libraries.py . fails to generate libraries


eyphka:
Running rosrun rosserial_arduino make_libraries.py . inside sketchbook/libraries as per http://wiki.ros.org/rosserial_arduino/Tutorials/Arduino%20IDE%20Setup results in missing libraries: 

**\* Warning, failed to generate libraries for the following packages: ***
    geometry_msgs
    sensor_msgs
    shape_msgs

mikeferguson:
Duplicate of #158

mikepurvis:
Fixed by #162.

Issue #158, opened 2015-01-29
Warning failed to generate libraries for the following packages: ***


rummanwaqar:
Running `rosrun rosserial_client make_libraries .` inside the `sketchbook/libraries` generates the following error:

Warning failed to generate libraries for the following packages: ***
  geometry_msgs
  sensor_msgs
  shape_msgs
  my_msgs

After a lot of testing, I found that the problem only occured if I used arrays (e.g. `float32[3] xyz`). Replacing it with three different fields (e.g. `float32 x`, `float32 y`, `float32 z`) fixed the problem.

_How do I use arrays in my messages and still make sure they compile?_

mikeferguson:
@rummanwaqar I presume you meant to run rosserial_arduino or similar package's make_libraries -- but I just did a check out and tested with rosserial_arduino and I can reproduce the error.

@mikepurvis this might be a major regression, previously the only packages that failed to generate were those with services with no return....

rummanwaqar:
@mikeferguson is there a previous version or something I can grab that works without this error?

mikepurvis:
Gracious, that's bad. Looking now.

The 0.6.3 release tag should be before the problem was introduced (almost certainly #152).

mikepurvis:
Fixed by #162.

PR #162, opened 2015-02-13
Fix string formatting mismatch introduced in #152.

mikepurvis:
I believe this corrects the issue reported in #158 and #161.


mikepurvis:
@rummanwaqar, @eyphka, can you guys try this out?

mitchellwills:
This works for me

PR #164, opened 2015-02-20
Make make_library.py more verbose re: *** Warning, failed to generate libraries...

chuck-h:
This diagnostic (copied from #162) helps resolve the mysterious **\* Warning, failed to generate libraries... failures. Saved my bacon, anyway!

PR #165, opened 2015-02-21
Make make_library.py error more verbose

chuck-h:
was: #164 (hydro-devel) for indigo-devel branch


mitchellwills:
Does it make sense to append the message to the failed string similar to how (missing dependency) gets appended?

mikepurvis:
LGTM, thanks.

@mitchellwills Can you clarify what you're meaning?

mitchellwills:
See #176

PR #166, opened 2015-02-25
Jade devel

ibaranov-cp:
New Jade branch... 

New 0xfc header, changing to uint32 data types.

Not complete yet, just in progress. Not required to merge, for @mikepurvis review ongoing work.


mikepurvis:
@ibaranov-cp Have you made any more progress here? I'd really like to set aside a block of time to merge stuff and put together the Jade release, but I don't want to until this work is done.

Looks like so far you've made the deserialization change for array lengths, but the trickier part of the problem is changing the protocol header itself, in order to accommodate longer _overall_ message lengths— this change needs to be coordinated across rosserial_client, rosserial_python, and rosserial_server.

Also, I really need to dig out the work I did a while ago building a unit testing framework for rosserial— it would help a lot with verifying this kind of thing. Perhaps I can look at it again this week.

Finally, note that you don't need to change any version numbers— that's handled by the catkin release pipeline, and in any case, Jade's first release will be 0.7.0.

ibaranov-cp:
Will work on it more, thanks for the tips!

mikepurvis:
Branch created independently.

Issue #143, opened 2014-09-24
rosserial arduino use PROGMEM in roslib for storing md5sum, message type, etc.


strothmw:
If programms on arduino run out of SRAM, the program may fail in unexpected ways; it will appear to upload successfully, but not run, or run strangely. The SRAM can be used up easily by storing a couple of constant strings in SRAM. 

This is critcal for ros_lib because it stores for each topic (publish/subscribe) 3 strings ( md5sum, message type name and topic name ). Particularly if I use topics with namespaces I have observed the problem on arduino uno after setting up only 3 or 4 topics.

Has anyone considered modifying the message generator for Ardunio such that it uses PROGMEM for storing md5sum, topic and stuff (http://www.arduino.cc/en/Reference/PROGMEM)? Is it possible at all?

The data is static and I think this would significantly reduce the SRAM usage....

See also:
http://answers.ros.org/question/193202/rosserial-arduino-use-progmem-in-ros_lib

mikepurvis:
The real trick is figuring out if there's a reasonable way to do it which still leaves the base rosserial_client package platform-agnostic, and ideally backwards compatible.

The low-hanging fruit would be to make it so that the `getMD5` and `getType` function writers in `make_library` are overridable on a per-platform basis, in much the same way that the serialization/deserialization code is to support float64 on 8-bit platforms. See here: https://github.com/ros-drivers/rosserial/blob/973ef48ac2be4b357cce6a2e22df3af3a42ec36c/rosserial_client/src/rosserial_client/make_library.py#L383-L388

Doing topic names is tougher— realistically that would be handled at the application layer, not by rosserial. The tricky part with that is that the `Publisher` and `Subscriber` classes expect to store a `const char*` of the topic name. They would have to learn to accept _functions_ which could populate these values into a supplied buffer, and then there's be a bunch of branching in the initialization code depending which interface was used.

mikepurvis:
For myself, I put a bunch of thought into this, and then spent a few dollars more to move to a 32bit MCU. If you're interested to work on this, please comment and I'll happily reopen the ticket so that the work can be tracked.

strothmw:
Hi Mike, thanks for your reply. Maybe I will try to dive deeper into this if I find some time in a couple of weeks. I will notify you in that case!

Thanks and regards

strothmw:
Hi Mike,
[...]

PR #168, opened 2015-02-27
 Pull request for: rosserial arduino use PROGMEM in roslib for storing md5sum, message type, etc. #143 

strothmw:
Pull request for solving the above issue (Using PROGMEM to reduce SRAM usage of ros_lib on Arduino)


strothmw:
Fixes #143

mikepurvis:
This intrigues me, and I appreciate the effort to document in the other ticket, but I'm still a little nervous about the amount of stuff being added.

How confident are you that the changes will not negatively impact non-serial users, non-Arduino users, and non-AVR users?

strothmw:
Are there any differences for running rosserial_emebedded_linux examples on a standard PC? If not maybe I can merge this with the jade devel branch for my fork an run the examples for embedded linux and the samples for rosserial_windowsand we can see if we will merge this into the jade devel branch.

For Arduino I've been using this very much during the last month, really notable more stable behaviour of the device!

AlexisTM:
Isn't it sufficient to use F macro (from http://playground.arduino.cc/Learning/Memory) ?
(I never used it and I am just wondering)

strothmw:
No, you can't. Because the original rosserial loginfo, logerror etc. functions of the nodehandle only accept const char* as input while the F macro "returns" a type const __FlashStringHelper * , i.e. you cannot pass it to the normal rosserial  loginfo, logerror etc. functions. The "Serial.println" of arduino is overloaded and has implementations for both,   const __FlashStringHelper *, const char* . Hence, you can use "Serial.println" with the F macro but not the normal rosserial loginfo, logerror etc. functions.

That's why I templated the methods in my fork (https://github.com/strothmw/rosserial/blob/indigo-devel/rosserial_client/src/ros_lib/ros/node_handle.h#L581) and use a base class ReadoutBuffer that by default just passes the const char* to the message function but for arduino is specialized to a FlashReadoutBuffer (https://github.com/strothmw/rosserial/blob/indigo-devel/rosserial_arduino/src/ros_lib/ros.h#L48) that still passes the const char* s but calls readFromFlash for the const __FlashStringHelper * s.

This allows using the loginfo, logerror, etc functions with both const char* strings and flash strings like here:
https://github.com/strothmw/rosserial/blob/indigo-devel/rosserial_arduino/src/ros_lib/examples/HelloWorldFlash/HelloWorldFlash.pde#L54

Passing the msg types / md5s  to the flash is simple, just requires overriding in the header generator for arduino, see https://github.com/strothmw/rosserial/blob/indigo-devel/rosserial_arduino/src/rosserial_arduino/make_libraries.py#L100. Here the PSTR macro assures the values stay in flash.

Unfortunately both PSTR and F only work inside function scopes. That's why for topics you need the PROGMEN keyword on a variable separat from the publisher object like here:
https://github.com/strothmw/rosserial/blob/indigo-devel/rosserial_arduino/src/ros_lib/examples/HelloWorldFlash/HelloWorldFlash.pde#L25

AlexisTM:
I appreciate your explanation!  This PR is awesome and is probably needed in some projects!

dkargin:
I've made an updated version of your pull request #295

Great job, it really allows us to make a more subscribers/publishers before RAM is completely exhausted on arduino (mega2560, mega328) side.

mikepurvis:
Closing in favour of #295.

Issue #169, opened 2015-03-05
Eliminate non-stdint.h types from Time, Duration classes


mikepurvis:
The `time.h` and `duration.h` headers should specify their member variables using the types defined in `stdint.h`. 

@tonybaltovski mentioned to me today that publishing a Duration from a 32-bit MCU back to the PC was causing a synchronization issue in his project.

tonybaltovski:
Thanks @mikepurvis.  In addition, is there a reason why a signed long is being using for in `ros/duration.h`  rather than unsigned? 
EDIT:  I just checked the ros duration message which is the same.

PaulBouchier:
ros time at http://wiki.ros.org/roscpp/Overview/Time defines that duration can be negative. It notes that ros::Time and ros::Duration have identical representations (signed), though obviously time can't be negative.

PaulBouchier:
As an interesting side note on the signed vs. unsigned issue: at work the coding standard calls for most variables like ints to be signed, even if the valid value range for the variable can never be negative (e.g time). The thinking is that if you use unsigned you're operating close to the boundary (0), where a simple mistake could cause roll-under, and you can't detect it with a <0 test. Conversely, if you use int for variables that should never be negative, it is an easily detectable error if they do become negative.
FYI FWIW

tonybaltovski:
Thanks @PaulBouchier, there seems to be case where this can be an issue with `time + duration = time` if the duration is negative but `normalizeSecNSecUnsigned()` will throw an error.  Good note nonetheless.

PaulBouchier:
Mike - can you check me on this - the logic in rosserial_client/src/ros_lib/time.cpp looks wrong:

  Time& Time::operator +=(const Duration &rhs)
  {
    sec += rhs.sec;
    nsec += rhs.nsec;
    normalizeSecNSec(sec, nsec);
    return *this; 
  }

[...]

PR #170, opened 2015-03-05
Eliminated non-stdint.h types from Time, Duration classes

tonybaltovski:
As stated in #169.  I believe I got them all.  This does fix my issue.

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -52,10 +52,10 @@
  * rosserial_python/src/rosserial_python/SerialClient.py
  */
 #define MODE_PROTOCOL_VER   1
-#define PROTOCOL_VER1		0xff // through groovy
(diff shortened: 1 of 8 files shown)


rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -99,13 +99,13 @@ namespace ros {
...
+      uint32_t rt_time;
 
       /* used for computing current time */
-      unsigned long sec_offset, nsec_offset;
+      uint32_t sec_offset, nsec_offset;
 
-      unsigned char message_in[INPUT_SIZE];
-      unsigned char message_out[OUTPUT_SIZE];
+      uint32_t message_in[INPUT_SIZE];
+      uint32_t message_out[OUTPUT_SIZE];
```
mikepurvis: These buffers are bytes, aren't they? I think this should be `uint8_t`?

tonybaltovski:
@mikepurvis, I just noticed the `client-stdint` branch so I will close this in favour of that branch.

PR #173, opened 2015-03-11
Bugfix for checksum.

attermann:
Publishing topics fails when messages are over 255 bytes in length due to checksum() function or'ing high and low byte instead of adding them.

### DIFF
--- a/rosserial_server/include/rosserial_server/session.h
+++ b/rosserial_server/include/rosserial_server/session.h
@@ -369,7 +369,7 @@ private:
   }
 
   static uint8_t checksum(uint16_t val) {
-    return (val >> 8) | val;
+    return (val >> 8) + val;
   }
 
   //// RECEIVED MESSAGE HANDLERS ////


mikepurvis:
I think there's something similar going on in #146. Did you find that no client library changes were necessary?

mikepurvis:
Oh wait, I see it. Yes, thanks for this contribution.

attermann:
I’m using the Arduino client, and no change was necessary.

BTW, it’s length >255 (not 256).

attermann:
No problem, glad I could contribute in a very small way.

I have been having some other issues with both the python and c++ servers (which I’ll detail elsewhere), but wanted to ask if this indigo-devel branch is still a work in progress, or should it be more or less complete?  I’m happy to contribute more as I plan to use rosserial extensively in my project.

Thanks for all the work you’ve done on rosserial BTW.  I find it’s an incredibly powerful piece of software for interfacing hardware to ROS.

mikepurvis:
The present _release_ is stable, but the indigo-devel branch is not. I'm in the middle of merging PRs and testing it out again, so definitely no guarantees.

Please do use the issue tracker to report other bugs or problems you've encountered.

@fergs gets the credit for most of rosserial's design and implementation. I'm happy to do what I can as far as maintaining it. :)

PR #174, opened 2015-03-11
Switch to stdint integers.

mikepurvis:
This allows the client to run on 64-bit systems.

Very similar to #170, from @tonybaltovski.


tonybaltovski:
Thanks @mikepurvis!

PR #176, opened 2015-03-11
Make error message more verbose and fail the build if message generation fails

mitchellwills:
This makes the end build output when generating messages a little more verbose so you can see what's wrong without scrolling through all of the output.
This also raises an exception in order to fail the build. The fact that the build continues even when message generation fails makes it pretty hard to figure out what's going on when things do fail and leads to some weird error messages when not all messages are generated (because apparently the regular c++ message headers are also in the include path when building rosserial client code)
I figured this was the easiest way to make it work in most scenarios, but it could also simply return an error code and then each script that calls rosserial_generate would call exit on the return value of the function. This is a change in behavior so I understand if failing the build isn't desired.

### DIFF
--- a/rosserial_client/src/rosserial_client/make_library.py
+++ b/rosserial_client/src/rosserial_client/make_library.py
@@ -42,6 +42,7 @@ import roslib.srvs
 import roslib.message
 import rospkg
 import rospy
+import traceback
 
 import os, sys, subprocess, re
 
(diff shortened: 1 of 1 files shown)


mikepurvis:
Historically we haven't failed the build over inability to generate messages, but I believe they should all now correctly build, which was not previously the case.

mitchellwills:
The only case I have seen recently where it fails (aside from the recently fixed message generation bug) is that I think the dependency resolution should be updated. Right now it looks like when it enumerates dependencies it ends up using build dependencies. This fails when a package is installed from a deb and there are build dependancies that are not listed as run depends. For example robot_localization build depends on gps_common but does not run depend on it (because all it uses is a header file in it). Not sure if it's correct/makes sense to switch to enumerating run_depends instead

mikepurvis:
Manually cherry-picked into `jade-devel`.

PR #183, opened 2015-05-20
Provide option to pass through CMake arguments.

mikepurvis:
In the CMAKE_COMMAND invocation. The use-case is primarily specifying additional paths to modules, for separately-packaged libraries.

(Will pull this into Indigo as well)


mikepurvis:
This is pretty clearly a non-breaking change; going to merge now.

PR #193, opened 2015-09-22
added WiFi support for Arduino clients

arnovanwouwe:
I have added the TCP socket connection method to the Arduino client part of your cool package.

Personally I have an ESP8266 chip -which has a specific WiFi arduino library- and have tested that it works for that. In theory it should work for other Arduino chips that use the refular WiFi.h library, but I have not been able to test it.

Quick user-guide on how to use it in the Arduino sketch: 
- "#include <WiFi.h>"
- "#define USE_WIFICON"
- bind to the access point with WiFi.setup(), as normal
- call nh.initNode(char\* hostname)

The initNode() will block until the TCP connection is created, retrying every 5 seconds.

The spinOnce() calls will detect broken TCP connections and try to restore the connection if it is broken (in a blocking way unfortunately). If it can not restore the connection immediately, it spinOnce() returns -3.

With the advent of the ESP8266 chip, I'm figuring that having WiFi capability in the Arduino part of your package is going to be in demand. But I'm  not really a big-time developer, and never use git so buyer beware =) If it looks interesting, feel free to include the changes.

Best regards, and thanks again for your wonderful package!

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -200,8 +200,17 @@ namespace ros {
         while( true )
         {
           int data = hardware_.read();
-          if( data < 0 )
+          if( data < 0 ) {
+#if defined(USE_WIFICON)
+		    if (!hardware_.isConnectionUp()) {		// check connection, if it is broken try to repair it N.B. should really add interrupt timer as this blocks !
+				configured_ = false;
+				return -3;
+			} else
+				break;
+#else
             break;
+#endif	    
+		  }
           checksum_ += data;
           if( mode_ == MODE_MESSAGE ){        /* message data being recieved */
             message_in[index_++] = data;
--- a/rosserial_arduino/src/ros_lib/ArduinoHardware.h
+++ b/rosserial_arduino/src/ros_lib/ArduinoHardware.h
@@ -50,6 +50,18 @@
 #elif defined(USE_USBCON)
   // Arduino Leonardo USB Serial Port
   #define SERIAL_CLASS Serial_
+#elif defined(USE_WIFICON) 
+  // connection to TCP socket
+  #if defined(ESP8266)
+	#include <ESP8266WiFi.h>
+  #else
+	#include <WiFi.h>
+  #endif
+  
+  #define SERIAL_CLASS WiFiClient
+  #if !defined(ROSSERIAL_PORTNUM)
+	#define ROSSERIAL_PORTNUM 11411					// default port number
+  #endif
 #else 
   #include <HardwareSerial.h>  // Arduino AVR
   #define SERIAL_CLASS HardwareSerial
@@ -66,7 +78,7 @@ class ArduinoHardware {
 #if defined(USBCON) and !(defined(USE_USBCON))
       /* Leonardo support */
       iostream = &Serial1;
-#else
+#elif !defined(USE_WIFICON)
       iostream = &Serial;
 #endif
       baud_ = 57600;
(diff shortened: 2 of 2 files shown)


ibaranov-cp:
Hey,

The newer esp8266 native Arduino driver will be merged in soon. That way, you don't need to have both arduino AND esp8266, just the latter.

https://github.com/esp8266/Arduino

arnovanwouwe:
hi @ibaranov-cp , thanks for your reply. 

This may seem a stupid question, but just want to make sure that I understand : you are saying that soon there will be a WiFi.h that works for the ESP8266, and that as a result we won't need to have the ugly "#if defined(ESP8266)" anymore?

It's just that just after that you say that we will only need "the latter", which means that we would only need the ESP8266 library. 

Did you mean to say "the first" instead of "the latter"? In that case indeed I'll wait for the patch to be merged and resubmit.

Thanks in advance =)

ibaranov-cp:
No, I mean that if you get the latest Arduino IDE, and use only the esp8266, you can natively run rosserial code on the esp8266 without any modifications (it uses the hardwareserial class for comms), I had this going just yesterday.

You can then do wifi connection, etc natively in the esp8266.

(Edit: Misunderstood original intent, everything is fine here)

ibaranov-cp:
Tested this with the esp8266 Mike, +1 from me!

Will have a seperate pull request for you soon with 2 examples, one using wifi to rosserial socket, the other using wifi to rosserial serial and webpage.

ibaranov-cp:
My only negative comment is that data seems to come through in buffered blocks, though if this is an issue with my wifi network, the computer, the esp8266, or the library, I have no idea.

rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -200,8 +200,17 @@ namespace ros {
...
           int data = hardware_.read();
-          if( data < 0 )
+          if( data < 0 ) {
+#if defined(USE_WIFICON)
+		    if (!hardware_.isConnectionUp()) {		// check connection, if it is broken try to repair it N.B. should really add interrupt timer as this blocks !
+				configured_ = false;
+				return -3;
+			} else
+				break;
+#else
```
mikepurvis, line 210: I'm not keen on having to add this housekeeping logic to rosserial_client. Is there a way to manage the wifi connection without requiring this modification?

ibaranov-cp, line 210: Not in any really easy way that I can see...

ibaranov-cp, line 210: Are you worried about code bloat or what?

The Wifi connection is fundamentally different than every single other instance of rosserial_arduino, as it doesn't use any serial (rs232) class, so can't get any of the nice hardware level managment.

mikepurvis, line 210: The client is abstract— all it worries about is sending and receiving characters. It manages for itself whether it is "connected" based on its time sync pings which go back and forth. So my objection is simply that there should be another way to manage this than adding more logic to the client.

The place for connection-managing logic is the Hardware class, IMO— you can either check it in the `read` method, or have something else which periodically checks the connection and attempts to recreate it as necessary.

vmatos:
@ibaranov-cp, take a look at my `rosserial_tivac_socket` and `rosserial_tivac_socket_tutorials` package. 
[rosserial_tivac_socket](https://github.com/vmatos/rosserial_tivac_socket)
[rosserial_tivac_socket_tutorials](https://github.com/vmatos/rosserial_tivac_socket_tutorials)

It uses `rosserial_client` without any modification for Ethernet TCP/IP rosserial socket server.
In the package you'll find an implementation for Arduino like and bare metal C/C++.

If there is an esp8266 arduino library, the hardware class can be as simple as this:
[TivaCHardware.h](https://github.com/vmatos/rosserial_tivac_socket/blob/master/src/ros_lib_energia/TivaCHardware.h)

mikepurvis:
Closing due to inactivity.

PR #194, opened 2015-09-23
Add support for HW Serial ports on the Teensy

garyservin:
ref: http://answers.ros.org/question/217986/rosserial-with-teensy-31-comm-port/


mikepurvis:
Looks harmless enough

PR #196, opened 2015-10-16
Fix for using different serial ports on mbed with rosserial

arlk:
Rosserial for Mbed defaults to using [USBTX, USBRX](https://developer.mbed.org/handbook/SerialPC). With these changes, you will be able to assign serial pins of your choice (eg: [p9, p10] ,[p13,p14] or [p28,p27] on [LPC1768](https://developer.mbed.org/platforms/mbed-LPC1768/)). This is super useful if you are using Xbees to setup the Mbed node on your network and have a lot of peripherals attached. 

The pin numbers can be easily defined when you instantiate the node handle:

``` C++
ros::NodeHandle  nh(p13,p14); //Your Xbee is connected at p13,p14
```

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -116,7 +116,11 @@ namespace ros {
        * Setup Functions
        */
     public:
+#ifdef ROS_MBED_HARDWARE_H_ 
+      NodeHandle_(PinName tx = USBTX, PinName rx = USBRX) : hardware_(tx, rx), configured_(false) {
+#else
       NodeHandle_() : configured_(false) {
+#endif
 
         for(unsigned int i=0; i< MAX_PUBLISHERS; i++)
 	   publishers[i] = 0;
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -116,11 +116,7 @@ namespace ros {
        * Setup Functions
        */
     public:
-#ifdef ROS_MBED_HARDWARE_H_ 
-      NodeHandle_(PinName tx = USBTX, PinName rx = USBRX) : hardware_(tx, rx), configured_(false) {
-#else
       NodeHandle_() : configured_(false) {
-#endif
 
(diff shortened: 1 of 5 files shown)


rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -116,7 +116,11 @@ namespace ros {
        * Setup Functions
        */
     public:
+#ifdef ROS_MBED_HARDWARE_H_ 
+      NodeHandle_(PinName tx = USBTX, PinName rx = USBRX) : hardware_(tx, rx), configured_(false) {
+#else
       NodeHandle_() : configured_(false) {
+#endif
```
mikepurvis: I don't care for modifying `rosserial_client` in a device-specific way like this. Can you put the logic in the MbedHardware class instead, perhaps as a method which can be called to set it up in the desired manner even after it has been constructed inside RosserialClient?

cf. https://github.com/ros-drivers/rosserial/blob/jade-devel/rosserial_mbed/src/ros_lib/MbedHardware.h#L17

arlk:
I may be missing a better method, but this was a quick and dirty fix. The user will now need to define macros for MBEDTX and MBEDRX to specify the pins over which they want to communicate.

mikepurvis:
Cool, thanks.

@garyservin This change okay with you?

garyservin:
Wouldn't it be easier to use the more OO way? At least since there's already a [constructor](https://github.com/ros-drivers/rosserial/blob/jade-devel/rosserial_mbed/src/ros_lib/MbedHardware.h#L17) for doing this.

[Here](http://answers.ros.org/question/198247/how-to-change-the-serial-port-in-the-rosserial-lib-for-the-arduino-side/?answer=198390#post-id-198390)'s how I do it for Arduino. And here is an example for mbed:

``` C++
class NewHardware : public MbedHardware
{
  public:
  NewHardware():MbedHardware(p9, p10, 57600){};
};

ros::NodeHandle_<NewHardware>  nh;
```

arlk:
@garyservin Yup, the OO way makes a lot more sense. Thanks. I shall close the PR.

PR #197, opened 2015-10-20
rosserial_mbed: moved from MODSERIAL to BufferedSerial library

romainreignier:
The [MODSERIAL](https://developer.mbed.org/users/Sissors/code/MODSERIAL/) library only supports some devices while [BufferedSerial](https://developer.mbed.org/users/sam_grove/code/BufferedSerial/) use the standard mbed Serial interface and seems to work on all mbed devices. Both provide a buffered serial communication using a ring buffer.
So I think that BufferedSerial should be enough as the MODSERIAL special features are not used by rosserial_mbed. The maintainer of the MODSERIAL repository used by rosserial_mbed often advise to use BufferedSerial :
- https://developer.mbed.org/questions/5147/Support-for-LPC1114/
- https://developer.mbed.org/questions/4179/Port-to-Nucleo-board/

I have tested the changes on some examples with both Nucleo F401RE and Nucleo F411RE. I have no NXP or Freescale based boards so @garyservin should try these changes first I think.


garyservin:
Hi! Thanks for the PR!

I agree that the BufferedSerial library would be the best option so we don't have to add support for each new board, but....

Before adding support for the STM32 Nucleo boars for MODSERIAL, I gave BufferedSerial a try. Most of the examples didn't work (or had a lot of lost packets) on my KL25Z, but worked fine with the MODSERIAL, hence I decided to go with it.

I'll run the tests again with the boards that I have (LPC1768, KL25Z and F401RE) with both libraries and make some kind of comparison between the two libraries. We can even define some sort of metrics so other people can test with other boards and report.

romainreignier:
Hi!
Actually, I have also noticed some lost packets on programs with a lot of traffic like the ADC example. But there were also some lost packets with MODSERIAL and by adding a small delay in the loop, the behavior were better.

As you said, maybe we could define some metrics.

garyservin:
Sorry for the excessive delay. 

@romainreignier I finally ran some tests with a more realistic robot (i.e. sending 9-DoF IMU, controlling motors, reading encoders, PID, etc) and both versions (BufferedSerial and MODSERIAL) seem to behave similarly. Also, as you've pointed out, it's easier to add support for new boards with the BufferedSerial library.

I see there are some merge conflicts from a previous PR I've sent. Would it be too much to ask you to update the PR so we can merge it?

Thanks, and again, sorry for the delay!

romainreignier:
Hi!
I did not have the occasion since then to try more the two libraries.
I have only merged with the upstream branch, deleted the files you have added and it seems to be enough.
Maybe you should try before merging.

garyservin:
@romainreignier I've been running more tests and everything seems to work well with the BufferedSerial library.

@mikepurvis it's OK to merge this PR. Thanks!

PR #200, opened 2015-11-16
Support for TI's TM4C123GXL and TM4C129XL evaluation boards.

vmatos:
Hello all!

We are proud to share our rosserial package for TI's launchpad boards.
This package allows you to use rosserial on TM4C123GXL and TM4C129XL evaluation boards, using:
- Energia IDE (arduino like)
- arm-none-eabi GNU toolchain through rosserial_client catking type build.

Documentation and tutorials are ready and can be found here:
[rosserial_tivac](http://wiki.ros.org/rosserial_tivac)
[Tutorials](http://wiki.ros.org/rosserial_tivac/Tutorials)

Let us know what you think. If it would be worth the inclusion of the package on rosserials' repo, tips, critics and comments.

EDIT: package versioning, repositories on package.xml and such details will need revision.

Cheers

### DIFF
--- a/rosserial_tivac/package.xml
+++ b/rosserial_tivac/package.xml
@@ -1,0 +1,27 @@
+<package>
+  <name>rosserial_tivac</name>
+  <version>0.0.1</version>
+  <description>
+  rosserial_tivac package provides the required hardware definitions for compiling rosserial_client targets for TivaC Launchpad 
+  evaluation boards.
+  </description>
+  <maintainer email="vmatos@robosavvy.com">Vitor Matos</maintainer>
+  <license>BSD</license>
+  <url type="website">http://wiki.ros.org/rosserial_tivac</url>
+  <url type="repository">https://github.com/robosavvy/rosserial</url>
+  <author>Vitor Matos</author>
+  <author email="vmatos@robosavvy.com">Vitor Matos</author>
+  
+  
+  <maintainer email="vmatos.pt@gmail.com">Vitor Matos</maintainer>
+  <license>BSD</license>
+  <url>http://ros.org/wiki/rosserial_tivac</url>
+
+  <buildtool_depend>catkin</buildtool_depend>
+
+  <run_depend>rospy</run_depend>
+  <run_depend>rosserial_msgs</run_depend>
+  <run_depend>rosserial_client</run_depend>
+</package>
+
+
--- a/rosserial_tivac/package.xml
+++ b/rosserial_tivac/package.xml
@@ -8,18 +8,12 @@
   <maintainer email="vmatos@robosavvy.com">Vitor Matos</maintainer>
   <license>BSD</license>
   <url type="website">http://wiki.ros.org/rosserial_tivac</url>
-  <url type="repository">https://github.com/robosavvy/rosserial</url>
+  <url type="repository">https://github.com/ros-drivers/rosserial</url>
   <author>Vitor Matos</author>
   <author email="vmatos@robosavvy.com">Vitor Matos</author>
-  
-  
-  <maintainer email="vmatos.pt@gmail.com">Vitor Matos</maintainer>
-  <license>BSD</license>
-  <url>http://ros.org/wiki/rosserial_tivac</url>
 
   <buildtool_depend>catkin</buildtool_depend>
-
-  <run_depend>rospy</run_depend>
+  
   <run_depend>rosserial_msgs</run_depend>
   <run_depend>rosserial_client</run_depend>
 </package>
--- a/rosserial_tivac/src/ros_lib/startup_gcc.c
+++ b/rosserial_tivac/src/ros_lib/startup_gcc.c
@@ -1,0 +1,490 @@
+/*
+ * Copyright (c) 2015, Robosavvy Ltd.
+ * All rights reserved.
+ * Author: Vitor Matos
+ * 
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the 
+ * following conditions are met:
+ * 
+ *   1. Redistributions of source code must retain the above copyright notice, this list of conditions and the 
+ * following disclaimer.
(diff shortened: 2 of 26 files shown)


rosserial_tivac/package.xml
```diff
@@ -0,0 +1,27 @@
...
+  <author email="vmatos@robosavvy.com">Vitor Matos</author>
+  
+  
+  <maintainer email="vmatos.pt@gmail.com">Vitor Matos</maintainer>
+  <license>BSD</license>
+  <url>http://ros.org/wiki/rosserial_tivac</url>
+
+  <buildtool_depend>catkin</buildtool_depend>
+
+  <run_depend>rospy</run_depend>
```
mikepurvis: Why does this depend on `rospy`?

rosserial_tivac/src/ros_lib/startup_gcc.c
```diff
@@ -0,0 +1,490 @@
+/*
+ * Copyright (c) 2015, Robosavvy Ltd.
```
mikepurvis: This looks like a TI-supplied file. Are you sure you're clear to assert copyright over it?

rosserial_tivac/src/rosserial_tivac/make_libraries
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python
```
mikepurvis: Let's not double this up— can you make this script simply call through to the other one?

rosserial_tivac/tivac-cmake/cmake/TivaCToolchain.cmake
```diff
@@ -0,0 +1,149 @@
+cmake_minimum_required(VERSION 2.8.3)
```
mikepurvis, line 45: Is this script your own concoction? If not, please add a header indicating where it came from and the circumstances under which it should be re-synchronized with upstream.

rosserial_tivac/package.xml
vmatos: Good question.
I based the development of this package on rosserial_arduino for good practices.
I guess the run depend is there for make_library script rospkg import.

rosserial_tivac/src/ros_lib/startup_gcc.c
vmatos: No I'm not clear about that. It's a case of copy pasties.
It's based on the TI-supplied Energia file https://github.com/energia/Energia/blob/master/hardware/lm4f/cores/lm4f/startup_gcc.c 
although several changes were made.

I'll retain the original header on the next commit.

rosserial_tivac/tivac-cmake/cmake/TivaCToolchain.cmake
vmatos, line 45: Yes it is.

rosserial_tivac/src/rosserial_tivac/make_libraries
vmatos: I would have to know if the script is called by `ros_client` cmake function `rosserial_generate_ros_lib` or by the user on the terminal to generate Energia's library.

I guess I can add an extra argument to the script, which can be assigned like this:

```
rosserial_generate_ros_lib(
  PACKAGE rosserial_tivac
  SCRIPT make_libraries.py <not_energia_argument>
)
```

The reason I divided into `ros_lib` and `ros_lib_energia` is because:
1. When I tried, Energia compiled all files that are copied into its library folders.
2. Probably cleaner maintenance.

mikepurvis: Ah, I didn't realise they were not the same. In that case, can you document (perhaps in the usage strings, which should be updated anyway) that the two are different because they are serving these slightly differing purposes?

vmatos:
Few changes and improvements:
- Documentation and usage cases.
- Changed naming for make_libraries scripts. Their role is hopefully more clear.

mikepurvis:
Cool, looks good. Thanks for the contribution.

rosserial_tivac/src/ros_lib/tivac_hardware.h
```diff
@@ -0,0 +1,226 @@
...
+      TivaCHardware::milliseconds = 0;
+      // Set up timer such that it produces one tick for each millisecond
+      SysTickIntRegister(TivaCHardware::SystickIntHandler);
+      MAP_SysTickPeriodSet(TivaCHardware::ui32SysClkFreq/SYSTICKHZ);
+      MAP_SysTickEnable();
+      MAP_SysTickIntEnable();
+
+      // Enable the peripherals for UART0
+      MAP_SysCtlPeripheralEnable(SYSCTL_PERIPH_UART0);
+      MAP_SysCtlPeripheralEnable(SYSCTL_PERIPH_GPIOA);
```
mikepurvis, line 109: It sucks about these pins being hard-coded. On the one hand, I don't want to make this file too much more complicated than it already is; certainly any serious user will just end up pulling this into their own project to customize it, and it does make sense to optimize for the dev board's configuration. On the other, it would be simple to have these as public members with default values set in the constructor, so that a user could trivially override them.

vmatos, line 109: ~~Which pins? A0 and A1 for UART0?~~

~~If you build your project with 'USB' on CMakeLists.txt, it will include a completely different file `tivac_hardware_usb.h'. Which does not use A0 and A1.~~

~~I implemented the hardware class on two separate files, UART and USB, instead of having one file with several ifdef sections.~~

vmatos, line 109: I see what you meant now. Hard-coded for UART0.

Yes, I think it will make things simpler having the code for the development boards only.
If some advanced user is able to design a shield/their own board, they'll be more than capable to clone and change up this initialization section.

PR #201, opened 2015-11-17
rosserial_mbed: Add support for Inforce's mbed PAC shield

garyservin:
This adds the necessary files to get Inforce's mbed PAC shield working with rosserial_mbed, the changes are only in the external MODSERIAL library, which we'll be unnecessary if we end up using the BufferedSerial library proposed in https://github.com/ros-drivers/rosserial/pull/197.


mikepurvis:
Cool, sounds good.

PR #202, opened 2015-12-01
[TIVAC] Improving Hardware class. No more error of multiple definitions

vmatos:
When developing a tutorial case for [FreeRTOS+rosserial_tivac](https://github.com/robosavvy/rosserial_tivac_tutorials/tree/master/freertos123) I encountered an error when someone includes ros.h on multiple files, gets multiple instantiations of the static members.

Solution:
Removing the variables which need instantiation for being used by the Class static functions, from the header onto its own source file.

### DIFF
--- a/rosserial_tivac/src/ros_lib/tivac_hardware.cpp
+++ b/rosserial_tivac/src/ros_lib/tivac_hardware.cpp
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2015, Robosavvy Ltd.
+ * All rights reserved.
+ * Author: Vitor Matos
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+ * following conditions are met:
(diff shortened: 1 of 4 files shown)


vmatos:
Don't merge just yet. My CppFu is failing me.
I would like your thoughts on how to best implement the interrupt functions within TivaCHardware class.

1. Have ```SystickIntHandler``` and ```UARTIntHandler``` as static member of TivaCHardware. Which requires some static member/global variables on ```tivac_hardware.cpp```
2. Have ```SystickIntHandler``` and ```UARTIntHandler``` as non-static member of TivaCHardware. Which requires pointer cast and gives out a pedantic warning.

What would be best? Everything within ```TivaCHardware``` class but with the pedantic warning and one less .cpp file. Or as it is?

mikepurvis:
Looks reasonable, thanks!

PR #203, opened 2015-12-13
Handle log messages in rosserial_server

mkrauter:
Addition to rosserial_server to handle incoming log messages from client (instead of the "Received message with unrecognized topicId (7)" warning)

### DIFF
--- a/rosserial_server/include/rosserial_server/session.h
+++ b/rosserial_server/include/rosserial_server/session.h
@@ -42,6 +42,7 @@
 
 #include <ros/ros.h>
 #include <rosserial_msgs/TopicInfo.h>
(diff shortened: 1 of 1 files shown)


bmagyar:
:+1: 
I've tested it with a Texas Instruments Tiva-C board, works like a charm!

@mikepurvis what do you think?

mikepurvis:
Agree, yes!

Any reason we wouldn't want this in indigo, too?

bmagyar:
Nothing at all I believe. This in fact implements the expected behaviour which is currently missing there too and as far as I know that codebase for the 2 is the same for rosserial.

PR #207, opened 2015-12-16
Generate unique node names

ThomasMatern:
Append the port to the node name, e.g. /serial_node_dev_ttyACM0 to allow running multiple instances on the same host.

### DIFF
--- a/rosserial_python/nodes/serial_node.py
+++ b/rosserial_python/nodes/serial_node.py
@@ -43,9 +43,6 @@ import sys
    
 if __name__=="__main__":
 
-    rospy.init_node("serial_node")
-    rospy.loginfo("ROS Serial Python Node")
-
     port_name = rospy.get_param('~port','/dev/ttyUSB0')
     baud = int(rospy.get_param('~baud','57600'))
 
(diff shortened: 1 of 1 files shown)


rosserial_python/nodes/serial_node.py
```diff
@@ -59,6 +56,9 @@
         port_name  = sys.argv[1]
     if len(sys.argv) == 3 :
         tcp_portnum = int(sys.argv[2])
+
+    rospy.init_node("serial_node{}".format(port_name.replace('/', '_')))
+    rospy.loginfo("ROS Serial Python Node")
```
mikepurvis, line 61: Are you able to access private params prior to calling `rospy.init_node`?

mikepurvis:
What is gained here over using `name="$(anon serial_node)"` in the roslaunch file?

ThomasMatern:
Sorry, I'm relatively new to ROS, just trying to find my way around.
I am running the serial_node directly from udev and pass the device name as a parameter, using rosrun directly, so there is no roslaunch file.

mikepurvis:
Cool, welcome!

Running a ros process directly from udev is certainly possible, but not typical or recommended. Some issues you're likely to encounter:
- Processes running from udev run as root, which you probably don't want for ROS.
- When something goes wrong, you'll be pretty blind— I don't believe udev captures stdout anywhere for processes it spawns.

The udev RUN capability is more about discovering names and such for devices; it's not meant to launch long-running processes.

Issue #175, opened 2015-03-11
rosserial_server service support


attermann:
I understand that development of service support is already underway in rosserial_server.  As it currently isn't working for me in either indigo release or devel, I wanted to find out what the status is before I delve in deeper.  I'm happy to help if it still needs more work.

For reference, I'm seeing warnings in rosserial_server like the following when running the ServiceServer example on Arduino...

[ WARN] [1426091785.412954964]: Received message with unrecognized topicId (2).
[ INFO] [1426091785.420138438]: waitForService: Service [/message_info] has not been advertised, waiting...
[ WARN] [1426091790.441098205]: Timed out waiting for message_info service to become available.

I assume that topicId (2) has something to do with the service advertised from Arduino.  I also noticed that rosserial_python doesn't complain about message_info service not being available, so is it safe to assume that Arduino is advertising this service properly, and it will work in rosserial_server once services are fully implemented?

mitchellwills:
This has nothing to do with a service advertised from the arduino. You also need to run rosserial_python/message_info_service.py when running rosserial_server
See: https://github.com/ros-drivers/rosserial/blob/indigo-devel/rosserial_server/launch/serial.launch
and http://wiki.ros.org/rosserial_server#Overview

attermann:
Thanks, I just came across that example launch file myself this morning.  It would be nice to not have any dependence on rosserial_python however.  And this leads back to my original question; what is the status of service implementation in rosserial_server?

mikepurvis:
I agree. One option would be to supply a list of message definitions to compile in, so that common message use cases wouldn't require the shim. If it were an extensible mechanism, platform-specific users could add in their own message definitions as well.

mitchellwills:
The issue with that is it requires rosserial_server to be built from source to use a new message type

attermann:
I am new to rosserial, and to ROS for that matter, so I don't have a very good understanding of message_info and service_info.  rosserial_python must be getting these definitions from somewhere at runtime though, so it doesn't seem too far fetched for rosserial-server to be able to retrieve them at runtime as well, perhaps by parsing msg and srv files into some internal structure. Could this work, or am I way off?

henoSH:
hello i tried also to use ServiceClient but i got the same error : waitForService: Service [/message_info] has not been advertised, waiting... 
have someone find a solution?

attermann:
[...]

PR #208, opened 2015-12-19
[rosserial_server] Support for service servers

mathias-luedtke:
Addresses #175 
Since this implementation uses condition variables I have added an abort_servers() member that aborts all running service calls in case of lost syncs and shutdown.

In addition I have added `require/service_clients` and `require/service_servers` handling.
To not break any existing installations lists that are not provided are skipped.


mikepurvis:
This looks great, thanks for the contribution. Going to merge as-is and then cherry pick it forward to jade-devel for the Kinetic release.

romainreignier:
@mikepurvis Is there a reason it was never merged in `jade-devel` in order to have it in Kinetic?

PR #210, opened 2016-01-04
rosserial_mbed: Add more sensor examples

garyservin:
Added more seeedstudio sensors and actuator examples, included:
- Grove Buzzer
- Grove Collision Sensor
- Grove Temperature and Humidity Sensor
- Grove PIR Motion Sensor
- Seeedstudio's Motor Shield

I've also added the Servo library for the Servo Example and added missing platforms to the button example makefile.


mikepurvis:
Looks good, though it's a mishmash style-wise. Are these from mbed's own examples and preserving mbed style? If not, please astyle them to match ROS C++:

http://wiki.ros.org/roslint#Tips

garyservin:
Thanks for the comment! I've applied the style fix to the examples.

mikepurvis:
LGTM, thanks!

PR #215, opened 2016-03-06
Backporting the #203 to indigo-devel

martonmiklos:
Backporting this change to indigo-devel:
https://github.com/ros-drivers/rosserial/pull/203


bmagyar:
Bump. Please resolve the conflict for this to be merged.

Btw: wouldn't it be simpler to cherry-pick the same commits to indigo-devel? (Just thinking out loud but I don't think the codebase has changed that much between the 2 versions.)

Issue #172, opened 2015-03-10
rosserial_server Session lifecycle


mitchellwills:
Is there a reason why the SerialSession destructor is private? This prevents it from being cleaned up properly when the program shuts down. The serial_node simply allocates a new one on the heap and throws away the pointer to it. Is this a mistake or is there a threading reason for it?

mikepurvis:
Sounds like a mistake. I would welcome a fix to this.

mitchellwills:
In digging into this further I am very confused by the expected lifecycle of SerialSession and Sessions in general. The session class seems to want to manage it's own lifecycle and delete itself when the connection fails and rely on the destructor to spawn a new instance of itself. This seems like a bad idea as it gives the user no control over the actual object and forces it to be allocated on the heap (in addition to causing a number of other issues: http://isocpp.org/wiki/faq/freestore-mgmt#delete-this)

mitchellwills:
This is also a memory leak

mikepurvis:
I agree that this architecture is non-ideal. In truth, it's a leftover from rosserial_server's origin as the boost [async_tcp_echo_server](http://www.boost.org/doc/libs/1_55_0/doc/html/boost_asio/example/cpp03/echo/async_tcp_echo_server.cpp) example, which uses this basic approach. The memory leak is extremely slow, so I'm not terribly concerned on that front, but it is still an indication of poor design.

I would be delighted to entertain proposals for how the architecture of rosserial_server might be rationalized and documented.

mitchellwills:
Wow i'm really surprised that's in an example. I've done a few projects recently that have been based around boost asio so I may take a look at potential ways to clean it up if your open to that. The biggest issue with the current architecture is that it gives you no way to interact with the session once it's created (explicitly shut it down, query state, etc)

mikepurvis:
Yes, I agree, and that's definitely a flaw. I suppose the best would be some kind of data structure which manages the session pointers, perhaps a vector or map, though it's not clear what you'd key them on. The main thing is being able to iterate them, and having a way to move that tortured session restart logic from within the session to an external manager process (ie, the session has a way to signal to the manager "shut me down now" or "shut me down and reopen me").

In any case, yes, please think on it and we'll discuss further soon.

mitchellwills:
I would initially propose something similar to [async_web_server_cpp](https://github.com/WPI-RAIL/async_web_server_cpp) where there is a server like object that creates sessions (like HTTPConnection) that handles the actual session.

PR #228, opened 2016-07-06
WIP: Explicit session lifecycle for the serial server.

mikepurvis:
This is a long overdue change which will resolve some crashes when USB serial devices return error states in the face or noise or other interruptions.

This is a work in progress— once support for the socket server is in place, this will represent a resolution to #172.

FYI: @mitchellwills @mikeodr

### DIFF
--- a/rosserial_server/include/rosserial_server/serial_session.h
+++ b/rosserial_server/include/rosserial_server/serial_session.h
@@ -49,49 +49,48 @@ class SerialSession : public Session<boost::asio::serial_port>
 {
 public:
   SerialSession(boost::asio::io_service& io_service, std::string port, int baud)
-    : Session(io_service), port_(port), baud_(baud), // io_service_(io_service),
-      timer_(io_service), interval_(boost::posix_time::milliseconds(1000))
+    : Session(io_service), port_(port), baud_(baud), timer_(io_service)
   {
-    connect_with_reconnection();
+    ROS_INFO_STREAM("rosserial_server session configured for " << port_ << " at " << baud << "bps.");
+
+    failed_connection_attempts_ = 0;
+    check_connection();
   }
 
 private:
-  ~SerialSession()
+  void check_connection()
   {
-    ROS_WARN("Serial session shutting down. Waiting 1 second for system state to settle.");
+    if (!is_active())
+    {
+      attempt_connection();
+    }
 
-    boost::shared_ptr<boost::asio::deadline_timer> timer
-          (new boost::asio::deadline_timer(
-              socket().get_io_service(),
-              boost::posix_time::seconds(1)));
-
-    // The timer instance is only passed to the callback in order to keep it alive for the
-    // required lifetime. When the callback completes, it goes out of scope and is destructed.
-    timer->async_wait(
-       boost::bind(&SerialSession::restart_session,
-                   boost::ref(socket().get_io_service()), port_, baud_, timer));
-  }
-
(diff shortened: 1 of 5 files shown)


rosserial_server/include/rosserial_server/serial_session.h
```diff
@@ -45,81 +45,63 @@
 namespace rosserial_server
 {
 
-class SerialSession : public Session<boost::asio::serial_port>
+class SerialSession
 {
 public:
   SerialSession(boost::asio::io_service& io_service, std::string port, int baud)
```
mikepurvis, line 51: Realistically, this is no longer a "Session", it's now a session manager. However, for the sake of not breaking API for other nodes which embed rosserial_server, it will remain as it is.

rosserial_server/include/rosserial_server/serial_session.h
```diff
@@ -45,81 +45,63 @@
...
-    timer->async_wait(
-       boost::bind(&SerialSession::restart_session,
-                   boost::ref(socket().get_io_service()), port_, baud_, timer));
+    ROS_WARN("Serial session shutting down.");
   }
 
-  static void restart_session(boost::asio::io_service& io_service, std::string port, int baud,
-                              boost::shared_ptr<boost::asio::deadline_timer>& timer)
+private:
+  void check_connection(const boost::system::error_code& ec)
```
mikeodr: `ec` doesn't actually seem to get used here.

mikepurvis: Agree, dropped.

mikepurvis:
Despite the large size of this, I don't think it's a terribly risky change, and resolves some significant crashes for us. Going to merge to jade-devel as-is, and then bloom for Kinetic. If there are no reported issues, then Jade will be bloomed eventually too.

rosserial_server/include/rosserial_server/session.h
```diff
@@ -180,9 +206,7 @@ class Session
     ros::serialization::IStream checksum_stream(stream.getData(), stream.getLength());
 
     uint8_t msg_checksum = checksum(checksum_stream) + checksum(topic_id);
-    if (client_version == PROTOCOL_VER1) {
-      msg_checksum += checksum(stream.getLength() - 1);
-    }
+    // msg_checksum += checksum(stream.getLength() - 1);
```
mikeodr: Remove?

rosserial_server/include/rosserial_server/session.h
```diff
@@ -83,34 +107,51 @@ class Session
...
-    PROTOCOL_MAX
-  };
-
 private:
+  /**
+   * Periodic function which handles calling ROS callbacks, executed on the same
+   * io_service thread to avoid a concurrency nightmare.
+   */
+  void ros_spin_timeout(const boost::system::error_code& error) {
+    ros_callback_queue_.callAvailable();
```
mikepurvis, line 144: This is the kernel of the change— periodically calling ROS callbacks on the ASIO thread, rather than from a separate one.

rosserial_server/include/rosserial_server/session.h
mikepurvis: Done.

mikepurvis:
Note that this change is fully backward compatible— the API for "embedding" rosserial_server is unchanged, it's simply that running `spin()` separately is now superfluous.

mikeodr:
LGTM other than previous fix-it comments.

PR #231, opened 2016-07-13
Implementation of native UDP rosserial server.

mikepurvis:
We use rosserial-over-UDP on Ridgeback, which is currently managed by a background socat instance which presents the UDP messages as a tty, which we then run the serial rosserial_server against (see [invocation here](https://github.com/ridgeback/ridgeback_robot/blob/97ba2d02b2972d83628dea79a19e10453255e7d0/ridgeback_bringup/scripts/can-udp-bridge.conf#L18)).

This change allows rosserial_server to do this natively, without depending on a secondary process.

@mikeodr @tonybaltovski


mikeodr:
LGTM.

tonybaltovski:
LGTM.  Can we backport to Indigo?

mikepurvis:
Eventually. Will test and soak for a while in Kinetic and Jade.

tonybaltovski:
Sounds great, thanks!

PR #233, opened 2016-07-18
Make the ~require param configurable from Session.

mikepurvis:
@mikeodr We're going to want this for our scenario, where a single robot_base node will be talking to two different MCUs with different sets of required topics.


mikeodr:
lgtm.

PR #234, opened 2016-07-20
Order packages by alpha rather than topologically.

mikepurvis:
There's no reason to have to build rosserial messages in topo order, and this squelches errors around packages you don't care about not being able to find their dependencies— the deps are either there and get messages, or aren't and don't.

Either way, if it's something you care about, you'll see it as a compile error in your firmware build, if it's not then it will silently fly under the radar.

FYI: @mikeodr @janapavlasek


mikeodr:
LGTM.

PR #236, opened 2016-08-05
Avoid runaway async condition when port is bad.

mikepurvis:
Failing to cancel the `require_check_timer_` can lead to a nasty situation where it fires after the session has stopped, which then immediately triggers another topics request and queues up another future check, despite that the session is stopped. The behaviour is that the event queue grows out of control until the server process consumes all available memory.

@mikeodr


mikeodr:
LGTM. Stop logic++

PR #240, opened 2016-08-11
Fix #109

### LINKED ISSUE #109: rosserial_client generated code assumes little-endianness
This happens to work out great for AVR and most ARM chips, but is not universal.

Potential steps:
- Introduce a flag in the library to specify which code to generate.
- Generate code which checks a compile flag
- Generate code which [checks at runtime](http://stackoverflow.com/questions/1001307/detecting-endianness-programmatically-in-a-c-program).
- Do nothing, but at least document it better somehow.

Any of the first three options is a potential mess to implement and verify—pretty hard to justify unless there's someone actively using a big-endian client with rosserial. In that case, the answer is probably documentation at least warning users. This ticket is a first step, and I've also added a note to the [rosserial_client wiki page](http://wiki.ros.org/rosserial_client).

IvanFeofanov:
Replaced the function memcpy on a universal (little-endian/big-endian) functions varToArr and arrToVar

### DIFF
--- a/rosserial_client/src/ros_lib/ros/msg.h
+++ b/rosserial_client/src/ros_lib/ros/msg.h
@@ -1,4 +1,4 @@
-/* 
+/*
  * Software License Agreement (BSD License)
  *
  * Copyright (c) 2011, Willow Garage, Inc.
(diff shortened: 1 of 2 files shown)


mikepurvis:
Thanks very much for looking at this! A few quick questions:
- Which platform(s) are you testing on?
- What is the intended meaning of the `varToArr` name? I find it kind of opaque; would you consider something like `toLittleEndian`/`fromLittleEndian`? Alternatively, http://stackoverflow.com/a/10346064/109517

IvanFeofanov:
1. Platforms: 
   - Atmega328
   - TP-Link TL-MR3020  (big-endian)
   - Linux x86-64 (little-endian)
2. varToArr means copying data from variable to array. This functions work with both big-endian and little-endian.

mikepurvis:
Just closed and reopened to get a travis test. Looks like there's an issue with `size_t` not being available?

mikepurvis:
Great, thanks!

PR #241, opened 2016-09-02
Add option for subscriber to call a member function

mikepurvis:
Also includes some simple tests, travis integration, and a README cleanup.

@mikeodr

### DIFF
--- a/rosserial_client/test/subscriber_test.cpp
+++ b/rosserial_client/test/subscriber_test.cpp
@@ -1,0 +1,55 @@
+#include "ros_lib/ros/subscriber.h"
+#include <gtest/gtest.h>
+
+
+bool callbackCalled;
+unsigned char buffer[1];
+
+class DummyMsg
+{
+public:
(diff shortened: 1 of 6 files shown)


rosserial_client/test/subscriber_test.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+  callbackCalled = false;
+  sub.callback(buffer);
+  ASSERT_TRUE(callbackCalled);
+}
+
+TEST(TestSubscriber, testMemberCallback)
+{
+  DummyClass cl;
+  ros::Subscriber<DummyMsg, DummyClass> sub("topic_name", &DummyClass::memberCallback, &cl);
```
mikepurvis, line 43: Types can't be inferred for constructors, only functions, but to do this as a function (eg, `NodeHandle::subscribe`), the Subscriber instance would have to be heap-allocated and passed around as a pointer, which is a more significant idealogical departure than this simple addition of a template override.

mikeodr:
Looks great!

PR #242, opened 2016-09-02
rosserial_python tcp server allowing socket address reuse

vmatos:
This PR is more about an annoyance of mine, than an actual bug.

When closing `serial_node.py` using TCP, I can't relaunch it right away due to the error `Python [Errno 98] Address already in use`.  Slightly annoying when `serial_node` is part of a bigger launch file where other nodes need restarting.

Basically Python respects TCP's TIME_WAIT (~60secs) for re-opening a new socket with the same address. Which is meant to allow live packets even after closing the connection. http://stackoverflow.com/questions/337115/setting-time-wait-tcp

I don't think that with rosserial the live packets would be a problem, because the protocol and the nodes are quite robust at dealing with out of place bytes from old streams. 
Which doesn't really happen when `serial_node.py` is correctly shutdown, in contrast with `serial_server` (cpp).

The other change, is what seems to me a little vestigial code. I don't seem to find the method `RosSerialServer.close()` used anywhere, and I don't think `RosSerialServer` even has a member `port`.


mikepurvis:
Looks good, thanks.

PR #243, opened 2016-09-03
Integration tests for rosserial

mikepurvis:
The idea here is to have some integration tests which exercise basic pathways of the rosserial server and client. As the test suite gets built up further, we can validate functions of the server and client portions, and be more confident merging improvements and refinements to the components.

This change has been a long time coming; I'm pleased to finally have it in what I believe to be a mergeable state (pending some style and other fixups, no doubt).

@PaulBouchier @mikeodr @vmatos @garyservin


mikeodr:
LGTM as long as all tests are passing.

PR #245, opened 2016-09-08
Fix AsyncReadBuffer for UDP socket case.

mikepurvis:
This should be a small CPU usage improvement for the TCP and serial use cases as well, since now each read() is grabbing as much as it can and buffering in-process, rather than making multiple small reads.

It's necessary for the UDP/datagram case because an individual read is fixed-size.

@mikeodr


mikeodr:
Code looks sane, although travis is unhappy.

mikeodr:
Good catch, LGTM.

PR #249, opened 2016-09-15
Adding updating nodehandle subscribe function.

mikeodr:
Nodehandle subscribe function did not support non-static
function for callbacks properly.

Also added related test.

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -348,11 +348,11 @@ namespace ros {
       }
 
       /* Register a new subscriber */
-      template<typename MsgT>
-      bool subscribe(Subscriber< MsgT> & s){
+      template<typename SubscriberT>
+      bool subscribe(SubscriberT& s){
         for(int i = 0; i < MAX_SUBSCRIBERS; i++){
           if(subscribers[i] == 0){ // empty slot
-            subscribers[i] = (Subscriber_*) &s;
+            subscribers[i] = static_cast<Subscriber_*>(&s);
             s.id_ = i+100;
             return true;
           }
@@ -366,7 +366,7 @@ namespace ros {
         bool v = advertise(srv.pub);
         for(int i = 0; i < MAX_SUBSCRIBERS; i++){
           if(subscribers[i] == 0){ // empty slot
-            subscribers[i] = (Subscriber_*) &srv;
+            subscribers[i] = static_cast<Subscriber_*>(&srv);
             srv.id_ = i+100;
             return v;
           }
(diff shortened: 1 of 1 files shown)


mikeodr:
@mikepurvis FYI.

rosserial_test/src/publish_subscribe.cpp
```diff
@@ -47,6 +47,28 @@ static void rosserial_string_cb(const rosserial::std_msgs::String& msg)
...
+    last_string_ = std::string(msg.data);
+  }
+
+  int getCount() { return rosserial_string_cb_count_; }
+  std::string getLastString() { return last_string_; }
+
+private:
+  int rosserial_string_cb_count_;
+  std::string last_string_;
+};
```
mikepurvis: May not be necessary, if we can just reuse the one from here: https://github.com/ros-drivers/rosserial/blob/jade-devel/rosserial_test/include/rosserial_test/helpers.h#L3

rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -348,8 +348,8 @@ namespace ros {
       }
 
       /* Register a new subscriber */
-      template<typename MsgT>
-      bool subscribe(Subscriber< MsgT> & s){
+      template<typename Subscriber>
+      bool subscribe(Subscriber& s){
```
mikepurvis: I know I suggested this, but given the other types which exist in this scope, `SubscriberT` might be a less confusing name.

rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -348,8 +348,8 @@ namespace ros {
...
         for(int i = 0; i < MAX_SUBSCRIBERS; i++){
           if(subscribers[i] == 0){ // empty slot
             subscribers[i] = (Subscriber_*) &s;
```
mikepurvis: While you're here anyway, please convert this to a proper C++ `static_cast<Subscriber_*>(&s)`.

rosserial_test/src/publish_subscribe.cpp
```diff
@@ -71,6 +93,32 @@ TEST_F(SingleClientFixture, single_subscribe) {
   EXPECT_EQ(string_msg.data, last_string);
 }
 
+/**
+ * Single message sent from a roscpp Publisher, received
+ * by a rosserial client subscriber.
+ */
+TEST_F(SingleClientFixture, single_subscribe_non_static) {
```
mikepurvis: Consider `single_subscribe_member_function` as the name of this.

rosserial_client/src/ros_lib/ros/node_handle.h
mikeodr: Got it, done.

rosserial_client/src/ros_lib/ros/node_handle.h
mikeodr: Done.

rosserial_test/src/publish_subscribe.cpp
mikeodr: Done.

mikepurvis:
Thanks, this was an oversight from the previous addition.

PR #253, opened 2016-10-27
Fix typo in serial error message

jonbinney:
`length`  and `bytes_remaining` are integers, and so don't have a `len()`.

Before this fix, when the wrong number of bytes were read over serial, an error about "`int` has no len()` was printed.


mikepurvis:
Excellent, thanks.

PR #257, opened 2016-11-29
Fixing message has no attribute _md5sum

### LINKED ISSUE #199: [Arduino] make_libraries: has no attribute _md5sum
**Note:** This could also be an issue for ros/actionlib or ros/geometry. They contain the messages causing the faulty behaviour. However, I am posting it here now as this is the common denominator. 

**edit: typo**
## Issue

Inside folder Arduino/libraries, running `rosrun rosserial_arduino make_libraries.py .` will cause some packages to fail with the error `('NoneType' object has no attribute '_md5sum')`.

The full backtrace:

``` python
*** Warning, failed to generate libraries for the following packages: ***
    tf ('NoneType' object has no attribute '_md5sum')
    actionlib ('NoneType' object has no attribute '_md5sum')
  File "/home/tgo/catkin_ws/src/rosserial/rosserial_arduino/src/rosserial_arduino/make_libraries.py", line 91, in <module>
    rosserial_generate(rospack, path+"/ros_lib", ROS_TO_EMBEDDED_TYPES)
  File "/home/tgo/catkin_ws/src/rosserial/rosserial_client/src/rosserial_client/make_library.py", line 584, in rosserial_generate
    raise Exception("Failed to generate libraries for: " + str(failed))
Exception: Failed to generate libraries for: ["tf ('NoneType' object has no attribute '_md5sum')", "actionlib ('NoneType' object has no attribute '_md5sum')"]
```
## Affects

_tf_: breaks on /opt/ros/share/tf/msg/tfMessage.msg
Version: 1.11.7-3 
_actionlib_: breaks on /opt/ros/share/actionlib/msg/TestResult.msg
Version: 1.11.3-3

I cannot verify whether other messages on these paths will break in a similar fashion, as the exception causes the whole package to stop generation.
## Expected result

All messages generated. In case of failure, a more descriptive error and/or fallback behaviour.
## Environment and reproduction

I am running ArchLinux (x86_64), with ROS Indigo instaled from the ArchLinux User Repositories. As rosserial-arduino is not in the AUR, it is installed from GitHub using the catkin method. My PYTHONPATH is set to Python2.7. As per the installation guide; I have also set my catkin_make to compile with the python 2.7 flags. 

I compile rosserial inside my workspace using `catkin_make`, and then source it with `source devel/setup.bash`. I then change to the Arduino/libraries folder, and attempt to generate the libraries. 

I have verified that the offending files exist on this path, and that they have the same permissions as the ones that do work.
## Attempted solutions
- Redownloading/compiling ROS-serial
- Reinstalling actionlib, tf
- Reinstalling catkin and related packages
- Reinstalling most of ros-indigo

mikeodr:
If a message package has issues with it being unable to be imported, message generation will fail with: `message has no attribute _md5sum`. Now skip messages that are unable to be imported with a error message to the build log.

This fixes #199.

### DIFF
--- a/rosserial_client/src/rosserial_client/make_library.py
+++ b/rosserial_client/src/rosserial_client/make_library.py
@@ -501,12 +501,17 @@ def MakeLibrary(package, output_path, rospack):
         sys.stdout.write('\n    ')
         for f in os.listdir(pkg_dir+"/msg"):
             if f.endswith(".msg"):
-                file = pkg_dir + "/msg/" + f
+                msg_file = pkg_dir + "/msg/" + f
                 # add to list of messages
                 print('%s,'%f[0:-4], end='')
-                definition = open(file).readlines()
-                md5sum = roslib.message.get_message_class(package+'/'+f[0:-4])._md5sum
-                messages.append( Message(f[0:-4], package, definition, md5sum) )
+                definition = open(msg_file).readlines()
+                msg_class = roslib.message.get_message_class(package+'/'+f[0:-4])
+                if msg_class:
+                    md5sum = msg_class._md5sum
+                    messages.append( Message(f[0:-4], package, definition, md5sum) )
+                else:
+                    err_msg = "Unable to build message: %s/%s\n" % (package, f[0:-4])
+                    sys.stderr.write(err_msg)
 
     # find the services in this package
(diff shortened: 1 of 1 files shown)


mikeodr:
@mikepurvis FYI.

mikepurvis:
Looks great, thanks.

PR #260, opened 2016-11-30
Updating rosserial_tivac for new TivaWare compilation flags

vmatos:
1. TivaWare 2.1.3.156 has changed its flags for hard float. Changed package's cmake accordingly.
2. Now we can specify the serial id for the flash target.


mikepurvis:
Looks good, thanks!

PR #263, opened 2016-12-06
adding custom baudrate on initNode

AlexisTM:
Hi,

This PR aims to add the ability to change the baudrate of the Serial Hardware on the initialisation of the nodeHandle.

Example to use the 115200 baudrate :

```
  nh.initNode(115200);
```

Tested and working on an Arduino, with `serial_node.py` on a baudrate of 115200 instead of 57600.

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -160,6 +160,26 @@ namespace ros {
         topic_ = 0;
       };
 
+      /* Start with a certain baudrate */
+      void initNode(long baudrate){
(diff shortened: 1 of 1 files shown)


mikepurvis:
My concern about this is that now every `Hardware` out there has to have a `setBaud` method, regardless of whether that makes any sense for them— for example, we have MCUs which speak TCP or UDP to the computer, for which it would be silly to have a `setBaud` method.

I think the correct way to handle this situation is to make your own `Hardware` class which expresses the details of how your hardware needs to be configured, including serial port setup, etc. I would welcome efforts to improve the documentation on this front.

PR #267, opened 2017-01-11
Fix typo in embedded linux when using library

kev-the-dev:
When trying to compile an embedded linux program using something similar to the makefile found [here](http://wiki.ros.org/rosserial_embeddedlinux/Tutorials/Advanced%20Techniques), I ran into some build errors on these lines. It seems the function signatures are different here then when used without libraries through embedded_linux_hardware.h

I also noticed that when messages with string enums are built, they don't put the quotes around the value, so they will not compile when imported. This is fixed with the second commit.


mikepurvis:
Looks good, thanks!

PR #268, opened 2017-01-31
Allows to specify chip's silicon revision for TivaWare.

vmatos:
CMake's function was extended, allowing the user to specify the board's chip revision.

When not specified, the default option is the same as in TivaWare examples.


mikepurvis:
Looks reasonable, thanks.

Issue #272, opened 2017-02-27
Arduino CMake crashes if source file is oddly formated


FirefoxMetzger:
In 

    rosserial/rosserial_arduino/arduino-cmake/cmake/Platform/arduino.cmake:919

each file is parsed for include commands line by line. As no policy is specified, CMake uses the old CMP0054, which means that it will "dereference variables and interpret keywords even if they are quoted or bracketed" (see [CMP0054](https://cmake.org/cmake/help/v3.1/policy/CMP0054.html)). 

This means the following will work:

    void foo(int bar)
    {
       // do nothing 
    }

Yet, the same code formated this way will break the generation:

    void foo
    (
        int bar
    )
    {
       // do nothing 
    }

because CMake interprets "(" , finds missmatching parenties and stops generation with errors.

This can be fixed using

    cmake_policy(SET CMP0054 NEW)

somewhere above line 919. This turns off further interpreting of quoted things in CMake If() statements, which in this case is the intended behavior, as now "(" will be matched against the regex, as intended.

I've created a fix and will send a pull request as well as link it here =)

FirefoxMetzger:
The fix has been approved and thus this bug does no longer exist.

PR #273, opened 2017-02-27
Fixed bug reported in Issue 272

FirefoxMetzger:
CMake will crash when interpreting source code in line 919, if the bracket "( )" span multiple lines.
This is because each line is scanned individually and the old CMP0054 is used. This causes CMAKE to try and interpret the line, find missmatching brackets and crash.

The fix changes to the new CMP0054, which resolves this problem.
This pull request has been submitted as a fix for #272


mikepurvis:
Looks good, thanks.

PR #278, opened 2017-03-18
Added capability to specify timeout in getParam methods

afrancescon:


afrancescon:
I added some convenient logs too in order to make it easy to understand what is going on.

mikepurvis:
Great, thanks!

PR #279, opened 2017-03-21
Rosserial_arduino: add Esp8266 support and an example

khancyr:
Hello,

This PR intend to provide a simple interface to use ROS with the popular ESP8266 with arduino support.
It use the ESP8266 as a client of a wifi AP where Rosmaster and rosserial socket server are launch.

I provide an hello world example to explain how to connect to the master.


mikepurvis:
Looks great, thanks for contributing this!

PR #281, opened 2017-04-22
Add support for STM32F1 with stm32duino.

romainreignier:
The cheap dev boards based on STM32 are becoming quite popular.

The Stm32duino community provides Arduino support for these boards ([wiki.stm32duino.com](http://wiki.stm32duino.com/index.php?title=Main_Page)).

So I have added a support for both USB and Hardware serial ports (with flag `USE_STM32_HW_SERIAL` like the one for Teensy).

Tested on Maple Mini and "Blue Pill" (cheap STM32F103C8T6 dev board) with
both USB Serial Port and Hardware Serial Port 1.


mikepurvis:
LGTM, thanks for the contribution!

(I'm a long-time user of rosserial with stm32, though I do it via [stm32plus](https://github.com/andysworkshop/stm32plus))

PR #288, opened 2017-05-20
Use ESP8266 header only if defined.

romainreignier:
#279 added support for ESP8266 but if the toolchain is not installed, the header file won't be find.

### DIFF
--- a/rosserial_arduino/src/ros_lib/ros.h
+++ b/rosserial_arduino/src/ros_lib/ros.h
@@ -37,7 +37,9 @@
 
 #include "ros/node_handle.h"
 #include "ArduinoHardware.h"
-#include "Esp8266Hardware.h"
+#if defined(ESP8266)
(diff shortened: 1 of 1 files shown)


tonybaltovski:
:+1:

mikepurvis:
Is this variable something that the user has to manually define, or it is supplied automatically with `Arduino.h`?

mikepurvis:
Okay, either way I agree with this change.

romainreignier:
@mikepurvis Actually, I do not know, I have used the same `#define` that @khancyr used [here](https://github.com/ros-drivers/rosserial/blob/be3898f90bf78737dcbbc13c69838591f7b0ee23/rosserial_arduino/src/ros_lib/ros.h#L52).

khancyr:
Yep nice change! Sorry about that I wasn't carefull enough. I think it is automagicly supply when you install ESP8266 support in arduino IDE

PR #291, opened 2017-06-05
Fix POLICY CMP0054 unknown

ValentinVERGEZ:
Hi, CMP0054 is only available since CMake 3.1.
I added a trivial condition to fix builds with rosserial_arduino on Ubuntu 14.04 (CMake 2.8.122).


mikepurvis:
Sweet, thanks!

PR #292, opened 2017-06-08
Added support for the Particle Photon

davidbsp:
This simple pull-request provides support for the Particle Photon via rosserial-arduino.

This is based on Gaurav Hirkelar's early work (https://github.com/ghirlekar/rosserial-libraries).


rosserial_arduino/src/ros_lib/PhotonHardware.h
```diff
@@ -0,0 +1,79 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef ROS_PHOTON_HARDWARE_H_
+#define ROS_PHOTON_HARDWARE_H_
+
+#include "Particle.h"
```
mikepurvis: Is this the only delta from `ArduinoHardware.h`? If so, I might prefer to handle this with an ifdef in that file rather than the whole file being duplicated.

davidbsp: I will look into that, and let you know. Thanks!

davidbsp:
@mikepurvis you were right. After some tests, here you have a "cleaner" commit. Thanks.

mikepurvis:
LGTM, thanks!

PR #293, opened 2017-06-13
Prevent time variable overflow leading to parameter timeout error

davidbsp:
This pull requests prevents a (wrong) timeout error that is thrown when fetching parameters from the parameter server, in situations where a board connects to a host running ROS more than 65,535 miliseconds (variable overflow) after the initNode() instruction.


mikepurvis:
LGTM, thanks!

PR #295, opened 2017-07-04
Updated pull request for: rosserial arduino use PROGMEM in roslib for storing md5sum, message type, etc. #143 #168

dkargin:
Updated pull request #168 and merged it with recent changes in jade-devel

rosserial-arduino memory footprint is greately reduced for mega2560, so we use it extensively in our projects.

### DIFF
--- a/rosserial_client/src/rosserial_client/make_library.py
+++ b/rosserial_client/src/rosserial_client/make_library.py
@@ -494,16 +494,17 @@ class Service:
 
 def MakeLibrary(package, output_path, rospack, target_specific_message_class, target_specific_service_class ):
     pkg_dir = rospack.get_path(package)
-    if target_specific_message_class is dict:
-        # Seems like we've got a dictionary with platform-specific type definitions
-        # Wrap it here into local message class
-        class LocalMessage:
+    if isinstance(target_specific_message_class, dict):
+        '''
+        Seems like we've got a dictionary with platform-specific type definitions
+        Wrap it here into local message class
+        We doing it to preserve original rosserial_generate API and ability to
+        customize message generation for specific platforms. 
+        '''
+        class LocalMessage(Message):
             ros_to_embedded_types_ = target_specific_message_class
 
-            def __init__(self):
-                pass
-
-        target_specific_message_class = LocalMessage()
+        target_specific_message_class = LocalMessage
 
 
     # find the messages in this package
--- a/rosserial_client/src/rosserial_client/make_library.py
+++ b/rosserial_client/src/rosserial_client/make_library.py
@@ -499,7 +499,7 @@ def MakeLibrary(package, output_path, rospack, target_specific_message_class, ta
         Seems like we've got a dictionary with platform-specific type definitions
         Wrap it here into local message class
         We doing it to preserve original rosserial_generate API and ability to
-        customize message generation for specific platforms. 
+        customize message generation for specific platforms.
         '''
         class LocalMessage(Message):
             ros_to_embedded_types_ = target_specific_message_class
@@ -564,35 +564,11 @@ def MakeLibrary(package, output_path, rospack, target_specific_message_class, ta
         msg.make_header(header)
         header.close()
 
-def get_dependency_sorted_package_list(rospack):
-    ''' Returns a list of package names, sorted by dependencies. '''
-    pkgs = rospack.list()
-    dependency_list = list()
-    failed = list()
-    for p in pkgs:
-        try:
-            depends = rospack.get_depends(p)
-            dependent = False
-            for i in range(len(dependency_list)):
-                if dependency_list[i] in depends:
-                    dependency_list.insert(i, p)
-                    dependent = True
-                    break
-            if not dependent:
-                dependency_list.append(p)
-        except rospkg.common.ResourceNotFound as e:
-            failed.append(p + " (missing dependency)")
-            print('[%s]: Unable to find dependency: %s. Messages cannot be built.\n'% (p, str(e)))
-    dependency_list.reverse()
-    return [dependency_list, failed]
 
-def rosserial_generate(rospack, path, target_specific_message_class, target_specific_service_class = Service ):
-    
-    # find and sort all packages
-    pkgs, failed = get_dependency_sorted_package_list(rospack)
-
+def rosserial_generate(rospack, path, target_specific_message_class, target_specific_service_class = Service):
     # gimme messages
-    for p in pkgs:
+    failed = []
+    for p in rospack.list():
         try:
             MakeLibrary(p, path, rospack, target_specific_message_class, target_specific_service_class )
         except Exception as e:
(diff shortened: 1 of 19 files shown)


dkargin:
Hm, need some time to make tests passing

mikepurvis:
As on the previous PR, my primary concern would be ensuring that none of the existing client APIs are broken by this change. It seems like the main issue is going to be in having to adapt the wrapper scripts that generate messages.

dkargin:
I guess we could make some sort of hack in rosserial_generate function. We could check whether 3rd parameter is Message subclass or dict (ROS_TO_EMBEDDED_TYPES table) and do appropriate wrappings inside. It will allow us to preserve original API for SDK generator. I'll try it next week

mikepurvis:
You don't want to overly complicate things, but if it's possible to do it without the whole thing becoming ridiculous, then yes, that would be my preference, and I'd be good to merge this at that point.

dkargin:
Added a hack. Still fighting with tests. I have kinetic and 'make tests' passes on my local computer. 
Travis fails with exceptions:

```
[rosconsole]: Unable to find dependency: cmake_modules
ROS path [0]=/opt/ros/indigo/share/ros
ROS path [1]=/home/travis/catkin_ws/src
ROS path [2]=/opt/ros/indigo/share
ROS path [3]=/opt/ros/indigo/stacks. Messages cannot be built.

[tf2_py]: Unable to find dependency: roslang
ROS path [0]=/opt/ros/indigo/share/ros
ROS path [1]=/home/travis/catkin_ws/src
ROS path [2]=/opt/ros/indigo/share
ROS path [3]=/opt/ros/indigo/stacks. Messages cannot be built.

[rosserial_client]: Unable to find dependency: angles
ROS path [0]=/opt/ros/indigo/share/ros
ROS path [1]=/home/travis/catkin_ws/src
ROS path [2]=/opt/ros/indigo/share
ROS path [3]=/opt/ros/indigo/stacks. Messages cannot be built.

[roswtf]: Unable to find dependency: cmake_modules
ROS path [0]=/opt/ros/indigo/share/ros
ROS path [1]=/home/travis/catkin_ws/src
ROS path [2]=/opt/ros/indigo/share
ROS path [3]=/opt/ros/indigo/stacks. Messages cannot be built.
```
and 
```
*** Warning, failed to generate libraries for the following packages: ***

    rosconsole (missing dependency)
    tf2_py (missing dependency)
    rosserial_client (missing dependency)
    roswtf (missing dependency)
```
causing subsequent tests fail as well, because include renaming in rosserial_test was skipped due to an exception

dkargin:
Should be ready now

strothmw:
Hi there,

thank you for keeping trace of the good old PR ;-) Good job making this ready for merge. 

I found that on Line 86 of rosserial_embeddedlinux/src/rosserial_embeddedlinux/make_libraries.py there is a copy-paste mistake in the original code and it should be:
'float64' :   ('double',            8, PrimitiveDataType, []),

there. AVR_Float64DataType is wrong there. The rest looks fine to me, and I also have the arduino and embedded linux clients of this in productive use since I created the other PR. I looking forward to coming back to the upstream ;-)

dkargin:
I've decided to be more conservative and restore original version of rosserial_embeddedlinux and rosserial_windows.

rosserial_client/src/rosserial_client/make_library.py
```diff
@@ -41,7 +41,7 @@
 import roslib.srvs
 import roslib.message
 import traceback
-
+import rospkg
```
mikepurvis: Please add this as an `exec_depend` to rosserial_client.

dkargin: Is it better just to remove 'import rospkg'? It was needed by the block of code that I've deleted later. Right now there are no direct usages of 'rospkg'.

mikepurvis: Drop it then.

mikepurvis:
Restoring the original versions for those instances which don't require the new API is great— it demonstrates that this change will be non disruptive to those who don't need it.

mikepurvis:
I don't understand the Travis failure— `python-catkin-pkg` is definitely installed. Going to re-kick and just see if it was a fluke.

dkargin:
What's about previous commit/build? Can you rerun it to check, whether it is caused by commit or some 'wrong moon phase' in travis?

mikepurvis:
Build for 80deb51b95d798e69880e95c6956bbfdcbeb9487 was previously fine, now re-running to be sure: https://travis-ci.org/ros-drivers/rosserial/builds/252785597

mikepurvis:
I wonder if `import rospkg` has some insane `sys.path` mutation logic going on in there?

mikepurvis:
Failing now for the previous commit too. So I think this is either an issue for Travis, or something that changed upstream. Either way, if the master branch is now broken too, then I think this can be merged as-is (since it previously built fine), and I'll deal with the build failure as a separate concern.

mikepurvis:
Thanks @dkargin and @strothmw for your work on this!

mikepurvis:
@dkargin We're getting endless "flash read not impl" spam in our logs now. I'm going to disable the message for now, and if there's something more subtle you'd like to do to fix it properly, you can submit a pull request for that.

strothmw:
Its an issue with the usage of the templated NodeHandle class. For Arduino we always need the FlashReadoutBuffer as last template argument. The default for the last template arg is the DefaultReaoutBuffer_ which only passes the strings and does not support the flash readout. This default is needed to keep the class signatures of the other rosserial client libs.

There are 2 potential causes:

One is that there are 2 new typedefs for the NodeHandle class for Arduino:

https://github.com/ros-drivers/rosserial/blob/jade-devel/rosserial_arduino/src/ros_lib/ros.h#L139

For Spark and ESP8266 (everything that uses the roslib generated by rosserial_arduino make_libraries.py) we also have to add explicit buffer sizes and FlashReadoutBuffer as last template arguments.

The other cause would be if you are directly using the NodeHandle_ template (rather than the typedef) in your application code. In this case you will also have to add the FlashReadoutBuffer as last template argument there. 

We should set up a fix PR for the typedefs in the Arduino roslib. Should we also somewhere in the docs discourage using the NodeHandle template directly and/or warn about explicitly setting the FlashReadoutBuffer if using it with Arduino?

mikepurvis:
@dkargin @strothmw Unfortunately, it's looking from our internal testing like merging this has broken the current devel branch— we have MCUs spewing log messages and consistently not subscribing to topics.

I haven't got the time to dig into a root cause, so I'm going to revert it for now, and if you'd like to resubmit it after testing for regressions, we can try again.

dkargin:
Hello Mike
Which MCU do you have and what's on the scene? Which host configuration do you use? Is host under some load?

mikepurvis:
The message is `Flash read not impl`, as discussed above in https://github.com/ros-drivers/rosserial/pull/295#issuecomment-317514414.

In order for this to be merged again, we will need a solution which does not require external rosserial ports to have an API breakage.

dkargin:
Hello Mike.
Can we make atmega-based rosserial clients use strings from flash directly, without template magic?
Could it be plausible solution for AVRs?

Memory constrains on AVRs are really hard, and odom+imu+joint_state publishers completely exaust RAM with no space for any control logic.

PR #298, opened 2017-07-18
Set PATH explicitly, to use system python for travis.

mikepurvis:


mikepurvis:
This was the failure that prevented #295 from passing CI before merge.

PR #307, opened 2017-08-04
Backporting support for subscriber to call member functions to indigo-devel

cmoser96:
Backporting the support for subscriber to call member functions from #241 to indigo-devel


mikepurvis:
Looks good, thanks.

PR #308, opened 2017-08-09
Add requestTopics to correct publish before configured

Baycken:
Right now, `serial_node.py` can not recover from `Tried to publish before configured` if the topic somehow was not configured properly.

This add a `requestTopics()` in order to recover from the above state.


mikepurvis:
Looks good, thanks!

PR #309, opened 2017-08-11
Backporting #248 to indigo-devel

pazeshun:
To suppress warn message `Serial Port read returned short (expected 99 bytes, received 64 instead).`
In #247 , this problem is reported only in ROS kinetic, but this also occurs in indigo.


mikepurvis:
No concerns with this reported on Kinetic, thanks!

pazeshun:
Thanks!

PR #312, opened 2017-09-07
In rosserial_arduino, changed embedded type size for ROS uint64 to 8 bytes from 4.

tom-f-oconnell:
As the title says. This fixed one error I was having, and it didn't seem to make sense to try to store a 64 bit datatype in 32 bits, as the generated serialization code was doing for my messages with fields of type uint64. I'm still getting oriented to rosserial, though, and it is possible I misunderstood something.

### DIFF
--- a/rosserial_arduino/src/rosserial_arduino/make_libraries.py
+++ b/rosserial_arduino/src/rosserial_arduino/make_libraries.py
@@ -60,7 +60,7 @@ ROS_TO_EMBEDDED_TYPES = {
     'int32'   :   ('int32_t',           4, PrimitiveDataType, []),
     'uint32'  :   ('uint32_t',          4, PrimitiveDataType, []),
     'int64'   :   ('int64_t',           8, PrimitiveDataType, []),
-    'uint64'  :   ('uint64_t',          4, PrimitiveDataType, []),
+    'uint64'  :   ('uint64_t',          8, PrimitiveDataType, []),
     'float32' :   ('float',             4, PrimitiveDataType, []),
     'float64' :   ('float',             4, AVR_Float64DataType, []),
     'time'    :   ('ros::Time',         8, TimeDataType, ['ros/time']),


mikepurvis:
On AVR 8-bit micros, the `double` type is aliased to a 32-bit float, so that's the history of why that one is the way it is. Is there a native int64/uint64 on those platforms, or are you on a modern Arduino that has a natively 32 bit chip?

romainreignier:
I think that @tom-f-oconnell is right.

In the avr-libc, the `uint64_t` is indeed a 8 bytes long integer (unless the `-mint8` options is used, which is not the case for Arduino) because it is a:

```C
typedef unsigned long long int uint64_t 
```
as shown [here](http://www.atmel.com/webdoc/avrlibcreferencemanual/group__avr__stdint_1gad27ed092432b64ff558d2254c278720f.html).

And in avr-gcc, a `long long` is 8 bytes long, as seen [here](https://gcc.gnu.org/wiki/avr-gcc).

So I have made a quick test, publishing a `std_msgs::UInt64` from an Arduino UNO:

```C++
#include <ros.h>
#include <std_msgs/UInt64.h>

ros::NodeHandle  nh;

std_msgs::UInt64 msg;
ros::Publisher chatter("chatter", &msg);

uint64_t hello = 0xdeadbeef112017;

void setup()
{
  nh.initNode();
  nh.advertise(chatter);
}

void loop()
{
  msg.data = hello;
  chatter.publish(&msg);
  nh.spinOnce();
  delay(1000);
}
```

And from the PC side, I get the following:

```bash
$ rostopic echo /chatter
data: 4010876951
---
```

Which correspond to the first 4 bytes:

````bash
$  python -c 'print 0xef112017'
4010876951
````

Finally, applying @tom-f-oconnell  patch, I get:

```bash
$ rostopic echo /chatter                                                                                         
data: 62678480396034071                                                                                                                                                               
--- 
```

That is right.

romainreignier:
@mikepurvis I think that PR can be merged.

romainreignier (approved):

mikepurvis:
Cool, thanks for digging into this, and sorry for the delay in reviewing!

PR #313, opened 2017-09-09
pyserial bug workaround to fix rosserial_test test against SerialClient.py

tom-f-oconnell:
With Ubuntu 16.04, ROS Kinetic, and (at least) either pyserial 3.0.1 or 3.4,  running `rostest rosserial_test rosserial_python_serial.test` fails with:
```
...
File "/usr/lib/python2.7/dist-packages/serial/serialposix.py", line 605, in _update_dtr_state
    fcntl.ioctl(self.fd, TIOCMBIS, TIOCM_DTR_str)
IOError: [Errno 22] Invalid argument
```
This adds an optional argument to `rosserial_python`'s `SerialClient.py` to fix this as a special case.

### DIFF
--- a/rosserial_python/nodes/serial_node.py
+++ b/rosserial_python/nodes/serial_node.py
@@ -51,6 +51,10 @@ if __name__=="__main__":
     port_name = rospy.get_param('~port','/dev/ttyUSB0')
     baud = int(rospy.get_param('~baud','57600'))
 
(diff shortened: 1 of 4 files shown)


tom-f-oconnell:
CI seems to have failed because the testing machine couldn't find cmake.  Could that be right?

mikepurvis:
Wow, that's quite a workaround. Thanks for figuring it out!

PR #316, opened 2017-09-22
Log recognized raw data

johnflux:
This is really useful when the device becomes disconnected and is sending non-ros information.


mikepurvis:
Cool, thanks for the contribution!

Can the data prints be wrapped in `repr` to encode any control characters? I'm also a bit concerned about flood control— could you check for the string being too long and cut it off at some reasonable max length?

PR #318, opened 2017-09-29
Use const in ros namespace instead of #define for constants. Fix #283

### LINKED ISSUE #283: MSG_TIMEOUT name conflict with ChibiOS
I am using ChibiOS to program STM32 microcontrollers and it already defines `MSG_TIMEOUT` that `rosserial` also defines [here](https://github.com/ros-drivers/rosserial/blob/jade-devel/rosserial_client/src/ros_lib/ros/node_handle.h#L69).

What would you think to rename it, maybe, `ROSSERIAL_MSG_TIMEOUT` to avoid any conflict because the current name is a bit too generic and may conflict with other libraries.

romainreignier:
As discussed, in issue #283 use constants within the `ros` namespace to avoid conflict with other library.
Note that I have also renamed a variable to avoid a conflict with a `#define` in ChibiOS.

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -42,32 +42,6 @@
 #include "rosserial_msgs/Log.h"
 #include "rosserial_msgs/RequestParam.h"
 
-#define SYNC_SECONDS        5
-
-#define MODE_FIRST_FF       0
-/*
- * The second sync byte is a protocol version. It's value is 0xff for the first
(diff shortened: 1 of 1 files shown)


mikepurvis:
Looks great, thanks!

Out of curiosity, why is the name change still necessary even after it is moved into the `ros` namespace?

romainreignier:
ChibiOS is included in my rosserial firmware and defines the same word `MSG_TIMEOUT`so the preprocessor replace that string in the `rosserial_client` code, breaking it. Now it is Ok.

PR #282, opened 2017-04-22
Add class member method callback support for Service Server.

romainreignier:
Based on the work done for the Subscriber (#241).

I did not manage to add test because there are conflicts with the full Ros version.

Here is a simple test for Arduino:

```C++
#include <ros.h>
#include <std_srvs/SetBool.h>

ros::NodeHandle  nh;
using std_srvs::SetBool;

class MyClass
{
  public:
    static void callback_class_static(const SetBool::Request& req, SetBool::Response& res)
    {
      res.success = true;
      if (req.data == true)
        res.message = "Class static method: true";
      else
        res.message = "Class static method: false";
    }

    void callback_class(const SetBool::Request& req, SetBool::Response& res)
    {
      res.success = true;
      if (req.data == true)
        res.message = "Class method: true";
      else
        res.message = "Class method: false";
    }
};

MyClass mc;

void callback_non_class(const SetBool::Request& req, SetBool::Response& res)
{
  res.success = true;
  if (req.data == true)
    res.message = "Non-class: true";
  else
    res.message = "Non-class: false";
}

ros::ServiceServer<SetBool::Request, SetBool::Response> server_non_class("non_class_srv", &callback_non_class);
ros::ServiceServer<SetBool::Request, SetBool::Response> server_class_static("class_static_srv", &MyClass::callback_class_static);
ros::ServiceServer<SetBool::Request, SetBool::Response, MyClass> server_class("class_srv", &MyClass::callback_class, &mc);

void setup()
{
  nh.initNode();
  nh.advertiseService(server_non_class);
  nh.advertiseService(server_class_static);
  nh.advertiseService(server_class);
}

void loop()
{
  nh.spinOnce();
  delay(10);
}
```

PR #321, opened 2017-09-30
[Arduino] Add an example to use Subscriber and ServiceServer in a class

romainreignier:
Following #282 I have added an example on how to use a Subscriber and a ServiceService as class members.

It was not documented before so there is a question on answers.ros.org about that subject: https://answers.ros.org/question/258249/publisher-or-subscriber-inside-a-class-rosserial/

### DIFF
--- a/rosserial_arduino/src/ros_lib/examples/BlinkerWithClass/BlinkerWithClass.ino
+++ b/rosserial_arduino/src/ros_lib/examples/BlinkerWithClass/BlinkerWithClass.ino
@@ -1,0 +1,70 @@
+#include <ros.h>
+#include <std_srvs/SetBool.h>
+#include <std_msgs/UInt16.h>
+
+ros::NodeHandle nh;
+
+class Blinker
+{
+public:
+  Blinker(byte pin, uint16_t period)
+  : pin_(pin), period_(period),
+    subscriber_("set_blink_period", &Blinker::set_period_callback, this),
+    service_server_("activate_blinker", &Blinker::service_callback, this)
+  {}
+
+  void init(ros::NodeHandle& nh)
+  {
+    pinMode(pin_, OUTPUT);
+    nh.subscribe(subscriber_);
+    nh.advertiseService(service_server_);
+  }
+
+  void run()
+  {
+    if(active_ && ((millis() - last_time_) >= period_))
+    {
+      last_time_ = millis();
+      digitalWrite(pin_, !digitalRead(pin_));
+    }
+  }
+
+  void set_period_callback(const std_msgs::UInt16& msg)
+  {
+    period_ = msg.data;
+  }
+
+  void service_callback(const std_srvs::SetBool::Request& req,
+                              std_srvs::SetBool::Response& res)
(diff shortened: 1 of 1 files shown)


rosserial_arduino/src/ros_lib/examples/BlinkerWithClass/BlinkerWithClass.ino
```diff
@@ -0,0 +1,70 @@
...
+
+ros::NodeHandle nh;
+
+class Blinker
+{
+public:
+  Blinker(byte pin, uint16_t period)
+  : pin_(pin), period_(period),
+    subscriber_("set_blink_period", &Blinker::set_period_callback, this),
+    service_server_("activate_blinker", &Blinker::service_callback, this)
```
mikepurvis, line 13: Might be slightly clearer to have these members as pointers and then initialize the objects in the constructor, but I'm not married to having it that way.

mikepurvis (approved):
One quick question about stack v. heap initialization, but overall this is great— thanks for contributing it!

rosserial_arduino/src/ros_lib/examples/BlinkerWithClass/BlinkerWithClass.ino
romainreignier: You mean, I should do dynamic memory allocation on a microcontroller? I am not a fan of that...

mikepurvis: It's a one-time allocation at startup, so that's pretty safe, but if you'd rather not, it can stay as-is. Another possibility would be adding do-nothing constructors to the classes so that they can be initialized empty and the reinitialized afterward.

romainreignier: I prefer the do-nothing constructor path. But I don't get why you want to initialize the class later? In the `setup()` 
function to avoid the `init()` method?

mikepurvis: All I'm suggesting is having the option for the example to look like this, which feels a bit more like idiomatic roscpp:

```
  Blinker(byte pin, uint16_t period) : pin_(pin), period_(period)
  {
    subscriber_ = ros::Subscriber("set_blink_period", &Blinker::set_period_callback, this),
    service_server_ = ros::ServiceServer("activate_blinker", &Blinker::service_callback, this)
  }

  void init(ros::NodeHandle& nh)
  {
    pinMode(pin_, OUTPUT);

    nh.subscribe(subscriber_);
    nh.advertiseService(service_server_);
  }
```

However, it's not going to work regardless without also re-specifying all the template arguments since there's no type inference on constructors, so that kind of blows it away anyway (and there'd need to be copy constructors added). So I think the best thing is to merge this contribution as-is, and if there's interest now or later in streamlining or adding to the API, that can be discussed on separate PR.

PR #322, opened 2017-09-30
[Arduino] Fix Esp8266 example directory

romainreignier:
An Arduino sketch must be in a directory which has the same name as the file in order to show up in the Arduino IDE.


mikepurvis:
Great, thanks!

PR #323, opened 2017-09-30
[Client] Fix a warning in comparison.

romainreignier:
The warning appears in the for-loop line 440 but the method `serialize()` returns an `int`.


mikepurvis:
LGTM.

PR #328, opened 2017-10-24
Changed hardcoded pin 13 to LED_BUILTIN

blubbi321:
To make the code work on an Arduino Nano V3 clone board.

(might need to be updated in BlinkM as well, but I didnt try that example)


romainreignier:
This change seems legitimate but I would also apply it to the other examples using a led:

- `ServoControl`
- `button_example`

About `BlinkM`, it actually uses a [BlinkM](http://thingm.com/products/blinkm/) module that is a i2c Led so no need to change anything there.

romainreignier (changes requested):
Apply the same on these examples:

    ServoControl
    button_example

blubbi321:
Thanks for the update!

PR #333, opened 2017-11-09
Fixing formatting on files.

mikeodr:
Astyle fixup.

Ran the following in the repo:
```bash
find -regextype egrep -regex '.*\.[ch](pp)?$' -exec astyle '{}' --style=allman --indent=spaces=2 --pad-oper --unpad-paren --pad-header --convert-tabs \;
```

### DIFF
--- a/rosserial_client/src/ros_lib/duration.cpp
+++ b/rosserial_client/src/ros_lib/duration.cpp
@@ -37,45 +37,47 @@
 
 namespace ros
 {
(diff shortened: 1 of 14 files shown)


mikepurvis:
I was picturing just doing `rosserial_client`— my main worry about the other repos is that there's some vendor code in there that probably shouldn't be restyled.

That said, maybe it does make sense to just fix everything?

mikeodr:
Will pair it back then.

mikepurvis:
Thanks, that is much more manageable.

PR #334, opened 2017-11-09
Add overall spin timeout to rosserial read.

mikeodr:
Reading may block for to long, while there is a check for a timeout, it is not at the global scope of the while(true) statement.

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -64,6 +64,10 @@ public:
 namespace ros
 {
 
+const int SPIN_OK = 0;
+const int SPIN_ERR = -1;
+const int SPIN_TIMEOUT = -2;
+
 const uint8_t SYNC_SECONDS  = 5;
 const uint8_t MODE_FIRST_FF = 0;
 /*
@@ -108,6 +112,9 @@ protected:
   /* used for computing current time */
   uint32_t sec_offset, nsec_offset;
 
+  /* Spinonce maximum work timeout */
+  uint32_t spin_timeout_;
+
   uint8_t message_in[INPUT_SIZE];
   uint8_t message_out[OUTPUT_SIZE];
 
@@ -139,6 +146,8 @@ public:
     req_param_resp.floats = NULL;
     req_param_resp.ints_length = 0;
     req_param_resp.ints = NULL;
+
+    spin_timeout_ = 0;
   }
 
   Hardware* getHardware()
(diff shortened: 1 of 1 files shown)


rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -209,6 +224,20 @@ class NodeHandle_ : public NodeHandleBase_
...
+      {
+        // If the maximum processing timeout has been exceeded, flush
+        // the receive buffer and exit with error.
+        // This is the maximum time that can be run for without caring
+        // what state the processing is in.
+        if ((hardware_.time() - c_time) > spin_timeout_)
+        {
+          // Reset back to first mode.
+          mode_ = MODE_FIRST_FF;
+          return -2;
```
mikepurvis: I supposed with a timeout of 0 (the default), the behaviour is unchanged, but it would be good to use a constant here instead, and also clarify that the expectation for users of `setSpinTimeout` is that they check for this return code on `spinOnce`, and if they see it, flush the buffer.

mikeodr: -2 was used for the other timeout below, I could make it something else also.

mikepurvis: I guess this actually works for a user even if they don't explicitly flush the queue— all it does is ensure that `spinOnce` only takes up to a certain amount of time, so if there's a burst of messages which arrive all at once, those will get processed over a _series_ of spin invocations, so long as there's the buffer space to accommodate them.

In any case, yes, please define SPIN_OK and SPIN_TIMEOUT consts and return them from this function in the appropriate spots.

rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -209,6 +230,20 @@ class NodeHandle_ : public NodeHandleBase_
...
+      if (spin_timeout_ > 0)
+      {
+        // If the maximum processing timeout has been exceeded, exit with error.
+        // The next spinOnce can continue off where it left off, or optionally
+        // based on the application in use, the hardware buffer could be flushed
+        // and start fresh.
+        if ((hardware_.time() - c_time) > spin_timeout_)
+        {
+          // Reset back to first mode.
+          mode_ = MODE_FIRST_FF;
```
mikeodr: @mikepurvis, if we intend to allow for just a continuation of processing, setting this would be detrimental. This kind of forces a buffer flush.

mikepurvis: Might be better to avoid doing so then. All the deserializing state is maintained between spins— if there's outside logic that drops portions of the buffer, the next spin will realise this and resynchronize anyway.

rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -64,6 +64,10 @@ class NodeHandleBase_
 namespace ros
 {
 
+const int SPIN_OK = 0;
+const int SPIN_ERR = -1;
+const int SPIN_TIMEOUT = -2;
```
mikepurvis, line 69: There's a case to be made that these should be `static const int` members of `NodeHandle`, but it's probably fine for them just to be here.

mikepurvis:
Looks like a great change, thanks for the contribution!

Issue #335, opened 2017-11-13
make_libraries.py fails for rosserial_embeddedlinux


jj-tetraquark:
After building and installing rosserial from source, running `rosrun rosserial_embeddedlinux make_libraries.py .` fails on account of the examples directory not being copied to the install directory in the catkin workspace:

```
Exporting to .
Traceback (most recent call last):
  File "/home/jdark/catkin_ws/install/lib/rosserial_embeddedlinux/make_libraries.py", line 89, in <module>
    shutil.copytree(rosserial_arduino_dir+"/src/examples", path+"/examples")
  File "/usr/lib/python2.7/shutil.py", line 171, in copytree
    names = os.listdir(src)
OSError: [Errno 2] No such file or directory: '/home/jdark/catkin_ws/install/share/rosserial_embeddedlinux/src/examples'

```

As well as not producing the expected examples directory, this also results in an incomplete ros_lib directory that is missing essential headers. Fixing this is a simple change in the CMakeLists.txt

mikepurvis:
Fixed via #336.

PR #336, opened 2017-11-13
Copy src/examples to install dir so make_libraries.py doesn't fail

jj-tetraquark:
src/examples for rosserial_embeddedlinux currently don't get copied to the install dir when running catkin_make causing make_libraries.py fail. Full details in issue #335 

Fix is simply adding src/examples to the directories that get copied in rosserial_embeddedlinux/CMakeLists.txt


mikepurvis:
Thanks!

Issue #237, opened 2016-08-07
Possible partial packet bug


SteveEasley:
The way `MSG_TIMEOUT` (https://github.com/ros-drivers/rosserial/blob/jade-devel/rosserial_client/src/ros_lib/ros/node_handle.h#L69) works seems to cause partial packets in the ring buffer to get dropped in most cases.

If a packet is only partially received into the buffer when spinOnce() is called, spinOnce() will properly return leaving the state of the process in tact. But once you call `nh->getHardware()->delay` (for times > 20ms), the subsequent call to spinOnce will drop the packet due to the 20ms timeout. See https://github.com/ros-drivers/rosserial/blob/jade-devel/rosserial_client/src/ros_lib/ros/node_handle.h#L69.

Was the idea that you should be delaying < 20ms? Seems like this could be implemented better, but before I offered a fix, I wanted to check if this was intentional.

hanophora:
I'm experiencing the same issue. Any packet too large to be sent in a single burst (around 60 bytes in my case if I remember well) will get dropped because of `delay` and processing time in the loop. I'm using rosserial_python and rosserial_arduino's ros_lib with an Arduino Mega.

I'm not sure what would be a good way to fix this. Maybe by using a unique packet header (a sequence of bytes that'd need to be escaped if seen in data) and constantly checking for it when receiving. That way truncated messages wouldn't mess everything up without the need for a timeout.

tom-f-oconnell:
I increased the `MSG_TIMEOUT` to 5000 (milliseconds), as a workaround, and it works OK for my application.  I also decreased the baud rate to 9600, because at baud rates >~20000 my messages were not reliably read out of the USB buffer before being overwritten.  This was for messages of beteween 100 - 300 bytes (usually in the middle of the range), using `rosserial_python` and an Arduino Mega.

Setting the baud rate consists of `nh.getHardware()->setBaud(9600);` **before** the call to `nh.initNode();`, and then adding an appropriate `baud` parameter when launching the `rosserial_python` or `rosserial_server` node.

hanophora:
I'm working on the problem right now, I should be done in the beginning of next week. I went ahead with my idea of changing the protocol to allow for detecting partial packets without any timeout.

PR #338, opened 2017-11-17
Change the protocol to avoid needing a timeout per message in node_handle.h

hanophora:
This solves #237 

## Technical description of changes

This PR changes the protocol from this:
```
│MAGI│VERS│  SIZE   │CHKS│  TOPIC  │   MESSAGE   │CHKS│
│ ff │ fe │    ┆    │    │    ┆    │    ┆...┆    │    │
```
to this:
```
│ MAGIC NUMBER │VERS│  SIZE   │CHKS│NULL│  TOPIC  │NULL│   MESSAGE   │NULL│CHKS│
│ ff ┆ ff ┆ ff │ fd │    ┆    │    │ 00 │    ┆    │ 00 │    ┆...┆    │ 00 │    │
```

In addition, the message is encoded as such:
```
[ff ff ff] is escaped as [ff ff 00 ff]
[ff ff 00] is escaped as [ff ff 00 00]
```

Overhead goes up 5 bytes + the cost of encoding.

The size field refers to the size of the *encoded* message. Checksum is not affected by encoding (only null bytes are added).

## Rationale

Using that protocol, the [ff ff ff] sequence always and only appears at the start of every packet. That means that encountering the sequence automatically resets the state machine.

Before those changes, partial packets were not nicely handled. ros_lib's NodeHandle kept a timer per message, and resetted the state machine on timeouts. The timeout was set very low (20ms), so that if a message could not be read in one burst, spinOnce returned and the user program would likely make it timeout.

SImply increasing the timeout worked, but I wasn't satisfied with this solution.

I'm not sure how frequent partial packets are, and how big an hassle they can be, but at least they can now be correctly handled without timeout. ros_lib's NodeHandle's state machine has been adapted and the timeout was dropped.

## Test

A testing architecture is available there : https://github.com/Pikrass/test-rosserial-fd

## Work still to be done

rosserial_python and rosserial_server still don't handle partial packets. Changes to do so seem non-trivial, since these programs are able to read multiple bytes at once and don't use a state machine. However, the protocol now allows it.

There's in fact a minor regression. Before this PR, an unsync'ed server could hope to "catch up" when waiting for the first ff of a message. Now, because it requires three, more messages are likely to be dropped until, by chance, the three ffs align with the read. (That is easier to fix though.)

### DIFF
--- a/rosserial_client/src/ros_lib/ros/msg.h
+++ b/rosserial_client/src/ros_lib/ros/msg.h
@@ -171,6 +171,10 @@ public:
         ffs = 0;
     }
 
+    // Return immediately if no escaping is needed
+    if (overhead == 0)
+	    return length;
+
     // Shift message to the right
     for (int i = length-1 ; i >= 0 ; --i)
       outbuffer[i+overhead] = outbuffer[i];
--- a/rosserial_client/src/ros_lib/ros/msg.h
+++ b/rosserial_client/src/ros_lib/ros/msg.h
@@ -141,6 +141,61 @@ public:
       var |= (arr[i] << (8 * i));
   }
 
+
+  /*
+   * @brief Serializes the message, then encodes it to eliminate sequences of
+   * three ff's.
+   *
+   * This function does its best to avoid allocating memory on the stack, at
+   * the expense of longer execution time. It does so by performing three
+   * passes. First, it computes how many more bytes will be needed. Then it
+   * shifts the message in the buffer that many bytes to the right. Finally it
+   * writes the message one last time inserting the escaping bytes.
+   */
+  int serializeAndEncode(unsigned char *outbuffer) const
+  {
+    int length = serialize(outbuffer);
+    int overhead = 0;
+    int ffs = 0;
+
+    // Compute overhead
+    for (int i = 0 ; i < length ; ++i) {
+      if (ffs == 2 && (outbuffer[i] == 0xff || outbuffer[i] == 0x00)) {
+        ffs = 0;
+        overhead++;
+      }
+
+      if (outbuffer[i] == 0xff)
+        ffs++;
+      else
+        ffs = 0;
+    }
+
+    // Shift message to the right
+    for (int i = length-1 ; i >= 0 ; --i)
+      outbuffer[i+overhead] = outbuffer[i];
+
+    // Write escaped message
+    ffs = 0;
+    unsigned char *w = outbuffer,
+                  *r = outbuffer + overhead;
+    for (int i = 0 ; i < length ; ++i, ++r) {
+      if (ffs == 2 && (*r == 0xff || *r == 0x00)) {
+        ffs = 0;
+        *w++ = 0x00;
+      }
+
+      if (*r == 0xff)
+        ffs++;
+      else
+        ffs = 0;
+
+      *w++ = *r;
+    }
+
+    return length + overhead;
+  }
+
 };
 
 }  // namespace ros
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -69,26 +69,36 @@ const int SPIN_ERR = -1;
 const int SPIN_TIMEOUT = -2;
 
 const uint8_t SYNC_SECONDS  = 5;
-const uint8_t MODE_FIRST_FF = 0;
+
 /*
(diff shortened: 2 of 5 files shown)


hanophora:
I can't reliably reproduce the failure on `rosserial_python_socket_test_publish_subscribe` (single_publish or single_subscribe). Looks like a timing issue, maybe a race condition. When I try to set watchpoints in gdb it always succeeds...

mikepurvis:
This is really cool— thanks for doing this work and offering the contribution. I'm still digesting the rationale and implications, but a few notes upfront:

- Because this changes the protocol, it would definitely have to wait for release until ROS Melodic.
- There would need to be full support in both `rosserial_python` and `rosserial_server`. If that effort is simplified by dropping support for previous versions of the protocol, I'm fine with that.

I think the biggest barrier on my side is persuading myself that there's no other way to solve the issue of incomplete packets within the context of the existing protocol (and in particular without resorting to byte stuffing). For example, a per-message timeout that adapts to the message length— which is given in the header, and reliable because it is protected by a checksum.

Finally, what are the contexts in which packets large enough to be affected by this deficiency are being sent? The most obvious case (mentioned in the previous ticket) is rosserial_windows; are there ways to resolve this specific to those use cases?

hanophora:
> There would need to be full support in both `rosserial_python` and `rosserial_server`. If that effort is simplified by dropping support for previous versions of the protocol, I'm fine with that.

Yeah, I implemented it without concern for compatibility between different versions of servers and clients. That's how it was currently done.

> For example, a per-message timeout that adapts to the message length— which is given in the header, and reliable because it is protected by a checksum.

The message length isn't really the main variable here. It has more to do with how often you call spinOnce. If there's too much delay between two calls (currently over 20ms), the timeout will expire. If you were in the middle of receiving a message when the function returned (data wasn't available immediately), you're screwed.

> Finally, what are the contexts in which packets large enough to be affected by this deficiency are being sent? The most obvious case (mentioned in the previous ticket) is rosserial_windows; are there ways to resolve this specific to those use cases?

 I have one use case myself: I send 280 byte long ROS messages. My calls to spinOnce are around 50 milliseconds apart. They consistently get dropped.

But again, the message length isn't what's important. It merely increases the chance it'll get dropped.
Let's say you transfer a 100-byte packet every second at 9600 bauds. The transfer takes 104 milliseconds (100/960). So the serial line is used one tenth of the time. If you call spinOnce during that time, it will receive a partial packet, and need to finish decoding on the next call. With the current timeout you'd need to call spinOnce every 18ms or so to allow it to do that, no matter the message length.


Here are other workarounds I can think of:

- Increase the per-message timeout.

It _would_ fully solve the issue for current Arduino hardware and libraries. Currently Arduino's input buffer for serial lines is 64 bytes for most hardware. That's 67 milliseconds at 9600 bauds (less for faster transfer rates). Setting the timeout to, say, 80 would ensure that, if the user calls spinOnce less than every 80ms, he would get dropped packets anyway, with or without per-message timeouts in rosserial, but just because the receive buffer gets full.

I'm not satisfied with this solution. It depends on variables we can't control. Maybe Arduino will increase the buffer some day, maybe for some future hardware. And what about non-Arduino hardware? I guess we don't want to drop support for embedded Linux. Also, a timeout too long would reduce the benefits it provides.

Also, rosserial could, in theory, prevent the client's buffer from overflowing by implementing flow control. That would be a neat feature. In fact I may think about implementing it myself, it would solve a problem I currently have.

- Implement an ack-based mechanism

Dropped messages would be noticed by the server and resent. However, the client still needs a way to know _when_ to drop a package. So even though some dropped messages would ultimately get received correctly (perhaps after _many_ tries), we're still stuck with a per-message timeout. If the user calls spinOnce too often, big messages will still consistently be dropped. In addition, this "solution" still requires changing the protocol.

- Completely remove support for partial packets

If we drop the timeout but don't change the protocol, what happens? The answer is: if there's a partial packet, a few more messages after it will be dropped, until the state machine gets in sync again. The timeout is there to prevent that and reset the state machine immediately, and that's why it needs to be quite short.

What are plausible causes for partial packets? There may be bugs in the server, but that's our fault, better fix it. There may be packets too big and too fast that overrun the client's buffer. That's the user's fault, he should fix it (I guess he needs those packets received anyway). Aside from that? I'm not sure...

So I don't think the consequences would be that bad. But of course I'd prefer a real solution.

PR #339, opened 2017-11-27
added support for arduino due

waspinator:


mikepurvis:
Can you explain the change a little bit? The snarl of ifdefs has gotten to be a little terrifying at this point.

(Actually the really ideal this would be a series of small unit tests which compile the file with different combinations of options to demonstrate that desirable settings are loaded in the supported cases... however, in the immediate, an explanation will suffice.)

rosserial_arduino/src/ros_lib/ArduinoHardware.h
```diff
@@ -49,14 +49,15 @@
...
+    #define SERIAL_CLASS Serial_
+  #else
+    #include <UARTClass.h>  // Arduino Due
+    #define SERIAL_CLASS UARTClass
+  #endif
 #elif defined(USE_USBCON)
   // Arduino Leonardo USB Serial Port
   #define SERIAL_CLASS Serial_
-#elif (defined(__STM32F1__) and !(defined(USE_STM32_HW_SERIAL))) or defined(SPARK) 
-  // Stm32duino Maple mini USB Serial Port
```
romainreignier, line 58: Why do you delete every STM32 entries?
You seem to use a version prior to https://github.com/ros-drivers/rosserial/commit/ce24e604db1cbab069332e183d42de4c5784d62f

PR #342, opened 2017-12-12
Retry opening the serial port every 3 seconds

hanophora:
This allows for connecting the device after launching ROS ans rosserial

### DIFF
--- a/rosserial_python/src/rosserial_python/SerialClient.py
+++ b/rosserial_python/src/rosserial_python/SerialClient.py
@@ -341,7 +341,7 @@ class SerialClient:
             self.port=port
         else:
             # open a specific port
-            while True:
+            while not rospy.is_shutdown():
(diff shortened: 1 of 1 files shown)


mikepurvis:
This is a great improvement, thank you! However, I believe that in order to exit cleanly you need to be check `ros.is_shutdown()` before you loop around again.

hanophora:
Right. There you go. :)

mikepurvis:
Cool, looks good. Note that if you're interested, a next-level version of this functionality is to also trap serial errors which occur _after_ that system starts up, and jump back into the initialization loop. This covers recovering after the client is unplugged, the USB port reinitializes, the MCU drops out for a minute while you flash it, etc.

`rosserial_server` does this more elaborate recovery loop; it would be great if `rosserial_python` did it too.

PR #345, opened 2018-01-01
Added ESP32 support

prnthp:
In similar fashion to adding support for the ESP8266.


mikepurvis:
Looks safe enough, thanks!

PR #346, opened 2018-01-03
[python] fix an unboundlocalerror

yoneken:
This line always failed with _UnboundLocalError: local variable 'msg' referenced before assignment_.
When _tryRead_ method at l.493 raises exception, the method always returns without any return value. 
 So, _msg_ at l.497 never be assigned.


mikepurvis:
Ha, good call!

PR #352, opened 2018-01-17
[stm32] Added STM32 support

yoneken:
I propose a rosserial_stm32 package.
Currently, this package only works on STM32F3xx series, but I hope anyone can port the codes to other STM32F series with minimum effort.


mikepurvis:
I'm generally supportive of this, but note that:

- STM32 has multiple interfacing libraries (officially Cube/HAL and the legacy stdperiph, plus third-party ones like libopencm3, stm32plus, ChibiOS's HAL)
- More complex micros like this have a lot more ways you might want to communicate with them: different UARTs, USB, TCP/IP, UDP/IP, etc.
- A micro like this is also more likely than Arduino to be hosting an RTOS, which could be FreeRTOS, ChibiOS, Nuttx, or something else. This in turn informs other aspects of how the interfacing would work (synchronous vs asynchronous, ISR use, DMA use, etc).

A package which represents itself as being general purpose for this family of micros needs to clearly call out which of these options/paradigms it is supporting and have a sane path to supporting more of them as required.

yoneken:
I totally agree.
This commit is just a first step. We have to keep developing to support more HALs, IDEs, peripherals, RTOSs, and also chips. Because [the development environment of STM32 keeps growing up!](https://www.eetimes.com/document.asp?doc_id=1332749)

romainreignier:
I agree with @mikepurvis I think that the STM32 world is a bit too broad to be included in this repo. Maybe you can host it in your Github account and add a link in the ROS Wiki.
I have myself a `rosserial_chibios` package that I have planned to release but on my own Github account.
What do you think?

yoneken:
Thank you for your kind reply. And sorry for my late reply.
I decided to create my own [rosserial_stm32](https://github.com/yoneken/rosserial_stm32) repository!

mikepurvis:
Yay, thanks for your contribution!

romainreignier:
@yoneken Great!

Maybe you should mention your repo in the [rosserial wiki page](http://wiki.ros.org/rosserial#Client_Libraries).

As a side note, you should explicitly mention that your work use the STM32CubeFx HAL from ST because there are many frameworks to program STM32 these days.

yoneken:
@romainreignier Thank you for your advice! I updated my Readme.

PR #355, opened 2018-01-29
Add support for boolean parameters

hanophora:
Trivial changes to support boolean parameters in rosserial_python and ros_lib.

Bools are simply passed as integers. That means if an user requests a parameter using a bool variable, but it turns out the param is really an int, this will convert it. Same thing the other way around. This behavior is different from, say, floats and ints, but I don't think it's an issue.


mikepurvis:
Looks great, thanks!

PR #362, opened 2018-03-22
Fix compiling on boost > 1.66

ProfFan:
This PR fixes the breaking changes on `async_*` brought by Boost 1.66.0.


mikepurvis:
Looks like this introduces an issue with `udp_socket_node`, per the Travis output. Can you investigate?

ProfFan:
Will check. Seems that I should use Boost-version-based check instead.

ProfFan:
@mikepurvis Fixed.

mikepurvis:
Great, thanks!

PR #364, opened 2018-03-23
Make rosserial Python 3 friendly

ProfFan:
Still testing. In my tests it works 25% of the time. The reason it does not work is under investigation.

### DIFF
--- a/rosserial_python/src/rosserial_python/SerialClient.py
+++ b/rosserial_python/src/rosserial_python/SerialClient.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+from __future__ import print_function
 
 #####################################################################
 # Software License Agreement (BSD License)
@@ -39,10 +40,18 @@ import roslib
 import rospy
 import imp
 
-import thread
+try:
+    import thread
+except ImportError as e:
+    import _thread as thread
+
 import multiprocessing
 from serial import *
-import StringIO
+
+try:
+    import StringIO
+except ImportError as e:
+    import io as StringIO
 
 from std_msgs.msg import Time
 from rosserial_msgs.msg import *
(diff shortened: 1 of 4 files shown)


rosserial_python/src/rosserial_python/SerialClient.py
```diff
@@ -56,6 +65,15 @@
...
 
+import sys
+if sys.version_info[0] >= 3:
+    def chr(a):
+        return bytes([a])
+    def ord(a):
+        if isinstance(a, int):
+            return a
+        else:
+            return int.from_bytes(a, 'little')
```
mikepurvis, line 76: Rather than defining these functions for Py3, I wonder if we could instead just adapt usage of them to use `struct` instead? https://docs.python.org/2/library/struct.html

ProfFan: Hmm. The solution in the PR is a shim for Python 2 functions just to MAKE IT WORK 😈 Actually I think the code needs a complete rewrite, too, however that would be a huge amount of work :) I am pretty happy to do the hard work, but we can merge this first to enable py3 usage.

ProfFan:
Also we can use [Kaitai](http://kaitai.io) as a code generation tool for all protocol stuff.

betab0t:
Fixes #386

Thrathra:
Hi, as python2 is now deprecated, is it possible to do a proper python3 support (I can contribute), maybe in another branch? Do we need to keep the python2 support?

mikepurvis:
@IThraen I have attempted it here, but it doesn't seem to be passing the tests, so there's clearly something not quite right about it: https://github.com/ros-drivers/rosserial/compare/py3-rosserial-python

Also, there are at least a few parts of that which will break on Py2, so if we don't want to figure out supporting 2 and 3 from the same code, we'll need to fork a noetic-devel branch for this.

Thrathra:
@mikepurvis I looked at the py3-rosserial-python branch, and it looks like the build fails because the rosserial_python/src/rosserial_python/SerialClient.py file imports "queue" but the python2 version of this module is named Queue. As I only have python3 on my machine I can't test this, but I think it's the correct way to fix this.

There's also runtime issues I am investigating now.

mikepurvis:
Thank you for the effort here. Going to close in favour of the approach in #480.

PR #378, opened 2018-07-02
rosserial for VEX Cortex

CanyonTurtle:
# rosserial for VEX Cortex

## Purpose
This additional package will allow the use of rosserial on the VEX Cortex, built on the PROS kernel.

VEX: https://www.vexrobotics.com/
PROS: https://pros.cs.purdue.edu/cortex/index.html

## Description of Changes
`/rosserial_client/src/rosserial_client/make_library.py` is modified to accept an additional optional paramater list in order to hack in a VEX include during message generation, only when running the VEX version of the build script.

e.g. the user will run `rosrun rosserial_cortex make_libraries.py .` and this script is almost exactly the same as the arduino version of `make_libraries.py`, except it is modified to pass an additional parameter into message generation to toggle (hardcoded) VEX includes. It is a quick hack, but it works. Library generation scripts in the other platform packages (e.g. arduino) still work the same with no modification to the source or command, since the optional parameter is simply unused in those scripts.

The package also includes a PROS project that is ready for use, when used with the `rosserial_cortex/README.md` instructions. 

## Future goals
In the future, I have a goal of finding a better design to injecting the VEX-specific includes, but for now, I think this modification is worth the extra platform it will provide.
Another future goal is to provide better documentation for this port of rosserial and its quirks. for example, avoid the global scope in PROS + rosserial projects, because it causes segmentation faults.


.gitignore
```diff
@@ -3,3 +3,4 @@
 rosserial_server/doc/
 *.out
 *.geany
+
```
tfoote, line 6: Try not to make unrelated whitespace changes. It clutters up the diffs and makes reviewing and merging much harder.

rosserial_cortex/.gitignore
```diff
@@ -0,0 +1,2 @@
+./working-rosserial-client/include/ros_lib/*
```
tfoote, line 1: These shouldn't be necessary. We should encourage people to work outside the source directory if possible.

rosserial_cortex/README.md
```diff
@@ -0,0 +1,56 @@
...
+2. An example PROS project, to get you started on using ROS with VEX.
+
+# Setup
+1. Install [ROS](http://wiki.ros.org/melodic/Installation/Source). This has been developed and tested on ROS melodic, but it should work on many earier/later versions as well.
+2. setup a workspace:
+  ```bash
+  source /opt/ros/melodic/setup.bash # or replace melodic with your corresponding ROS version name
+  mkdir -p <your-workspace-name>/src
+  cd <your-workspace-name>/src
+  git clone https://github.com/CanyonTurtle/rosserial.git
```
tfoote, line 18: This link will be out of date after merging.

rosserial_cortex/nodes/serial_node.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python
```
tfoote, line 1: Is there any difference between this and the rosserial_arduino version? https://github.com/ros-drivers/rosserial/blob/jade-devel/rosserial_arduino/nodes/serial_node.py#L1 if not we should reuse it. And if so look at parameterizing it for reuse.

tfoote (commented):
Overall this generally all looks good to bring in. But this pull request is too large and doing several things at once to effectively review. 

This should probably be broken up into two or three PRs for separate review.

Also there appear to be a bunch of binary files and possibly generated files included in these commits. Lets talk tomorrow about how to refactor this for easier review. 

I put in a few inline comments but with the size of the diff and complexity I couldn't really get through it all.

romainreignier:
I second @tfoote on most comments about the PR.

But I also want to question the name of this new package. Cortex is a common name in the microcontroller world because it describes an architecture family (ARM).
So I would suggest to change the name to one that reflect the VEX membership and not too ambiguous with the ARM Cortex-M family.

Also, as you mentioned, it would be better to move the VEX customizations to the package and not alter the generic `rosserial_client`.

CanyonTurtle:
See #380, I am closing this one and re-trying in a separate PR. Thanks for the suggestions!

PR #379, opened 2018-07-03
Enhance ArduinoTcpHardware

bochen87:
- Add support to check for current TCP connection status
- stop tcp before re-connecting (sometimes it fails to reconnect and hangs indefinitely and stopping the tcp first helps with this scenario)

### DIFF
--- a/rosserial_arduino/src/ros_lib/ArduinoTcpHardware.h
+++ b/rosserial_arduino/src/ros_lib/ArduinoTcpHardware.h
@@ -68,6 +68,10 @@ public:
 
   void init()
   {
+    if(tcp_.connected())
+    {
+      tcp_.stop();
+    }
     tcp_.connect(server_, serverPort_);
   }
 
@@ -78,6 +82,7 @@ public:
     }
     else
     {
+      tcp_.stop();
       tcp_.connect(server_, serverPort_);
     }
     return -1;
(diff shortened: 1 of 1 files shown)


romainreignier:
I did not try the changes but they seems legitimate.

But I am wondering how do you use that code? Because the hardware's `init()` method is only called from the `NodeHandle::initNode()` which should only be called once per program.

bochen87:
The node handle has a hardware object - you can access it by calling the getHardware() function https://github.com/ros-drivers/rosserial/blob/fe1e233f39d56931b5e223c5e53a2d9c2f9fb664/rosserial_client/src/ros_lib/ros/node_handle.h#L153

romainreignier:
@bochen87 Ok, that is what I thought. So sometime in your firmware, you detect that there is an issue with the connection and you re-call the `init()` method?

bochen87:
exactly, In my main loop every few iterations i check the conn status and reconnect if needed.

romainreignier:
@bochen87 Ok, I see. Interesting.

I have never tried the ArduinoTcpHardware in a long run but if it useful to `init()` we should add it to the examples.

romainreignier (approved):

bochen87:
Since changes are approved, can somebody merge pls?

mikepurvis:
Sorry for the delay!

PR #388, opened 2018-09-04
Indigo devel

craterghost:


mikepurvis:
Looks like this was created in error?

craterghost:
I'm sorry I couldn't make it earlier... No it wasn't a kernel error, the error is also being showed due to the most ridiculous reasons like connecting wrong ports or trying to communicate with wrong baud rates. So I don't really remember what it was in this case but the final result is the knowledge of this error being able to mean basically everything. 🌈 Still thanks a lot guys✌🏼

PR #393, opened 2018-09-19
fix no attribute message_cache issue in message_info_service

dlguo-cpr:
Reorder the creation of attributes and services to prevent the _message_info_cb function is triggered before the message_cache attribute is created.


mikepurvis:
Great, thanks!

PR #414, opened 2019-02-26
rosserial_python: handle service responses correctly

### LINKED ISSUE #408: rosserial_python : Service handler returned None
I've setup service server on stm32 :

```
ros::NodeHandle nh;

void srv_callback(const custom_srv::AddTwoIntsRequest &req, custom_srv::AddTwoIntsResponse &res){
	res.Sum = 2;
}

ros::ServiceServer<custom_srv::AddTwoIntsRequest,custom_srv::AddTwoIntsResponse> server("add_two_ints",&srv_callback);

void setup(){
	nh.initNode();
	nh.advertiseService(server);
}
void loop(void)
{
	nh.spinOnce();
}

```
But when I try to call the service from rosserial_python error occurs :
```
$rosservice call /add_two_ints "A: 0
B: 0" 
ERROR: service [/add_two_ints] responded with an error: service cannot process request: service handler returned None

```

I tried to debug this error on stm32. Apparently, the error occurs befor computing srv_callback(). I assume that this is related with rosserial_python package but it's onyl guessing.

lenzc:
`self.parent.send()` always returns None. Therefore we never waited for the service response and receive an error message `"ERROR: service [/topic] responded with an error: service cannot process request: service handler returned None"`

This PR also adds a small wait to reduce cpu load during waiting for the response.
This fixes #408.

### DIFF
--- a/rosserial_python/src/rosserial_python/SerialClient.py
+++ b/rosserial_python/src/rosserial_python/SerialClient.py
@@ -170,9 +170,9 @@ class ServiceServer:
         data_buffer = StringIO.StringIO()
         req.serialize(data_buffer)
         self.response = None
-        if self.parent.send(self.id, data_buffer.getvalue()) >= 0:
-            while self.response is None:
-                pass
+        self.parent.send(self.id, data_buffer.getvalue())
+        while self.response is None:
+            time.sleep(0.001)
         return self.response
 
     def handlePacket(self, data):
--- a/rosserial_python/src/rosserial_python/SerialClient.py
+++ b/rosserial_python/src/rosserial_python/SerialClient.py
@@ -170,8 +170,11 @@ class ServiceServer:
         data_buffer = StringIO.StringIO()
         req.serialize(data_buffer)
         self.response = None
+        self.error = False
         self.parent.send(self.id, data_buffer.getvalue())
         while self.response is None:
+            if self.error:
+                raise RuntimeError("Service call failed.")
             time.sleep(0.001)
         return self.response
 
@@ -780,7 +783,8 @@ class SerialClient(object):
                     try:
                         if isinstance(data, tuple):
                             topic, msg = data
-                            self._send(topic, msg)
+                            if self._send(topic, msg) < 0:
+                                self.error = True
                         elif isinstance(data, basestring):
                             self._write(data)
                         else:


chrisspen:
It looks like you identified a definite problem. When I refactored the code to make it asynchronous, I didn't update that one check for the return value from `send()`. The only remaining problem I see is an infinite loop might still happen if someone attempts to send a message larger than then the buffer, causing [this code](https://github.com/ros-drivers/rosserial/blob/5e4e6db497b89f752c792bda5fddf1cadd1a79f0/rosserial_python/src/rosserial_python/SerialClient.py#L757) to never send the message, resulting in a response never being received but no error or exception ever being shown.

Would it make sense to add a timeout to `ServiceServer.callback()` to handle those edge cases?

lenzc:
You are right, I've overseen this corner case. I added an error flag set by `_send()` notifying the `ServiceServer.callback()`. Does this solve the problem?

doisyg:
Any update on this ? It would be really useful to have a released fix

altineller:
@lenzc I have carefully patched my system, and no, this does not solve the problem.

But I have carefully studied the problem comparatively both on a ubuntu16/kinetic and ubuntu16/melodic workstation, and I have some further input and observations on the problem.

1. This problem not only affects rosserial_arduino, but also rosserial_tivac. (http://wiki.ros.org/rosserial_tivac)

2. It is not that the service fails or the SerialClient.py can not handle a service error. The service response is not processed, or the message is wrongly deserialized.

3. First, I have uploaded to my TivaC board the code at http://wiki.ros.org/rosserial_tivac/Tutorials/RGB%20srv - using ubuntu16/kinetic. When this service is called, it adjust the rgb led on the board, and returns resp.result = true. While on kinetic, the response is acquired correctly, in ubuntu18/melodic it is not. It will return "ERROR: service [/led] responded with an error: service cannot process request: service handler returned None"

4. The service handler returns none, indicating an error, but indeed the service works fine, and you can adjust the led colors. It just returns none from the service handler. (I have also tested this with a print statement, self.parent.send(self.id, data_buffer.getvalue()) returns None.

5. After hopelessly debugging the problem, in the end, I have replaced the SerialClient.py at my melodic install the SerialClient.py from kinetic install. And it worked! I was able to get service response correctly as True.

6. I think the issue is related to processWriteQueue which was introduced in the melodic version. while send() directly sends in the previous version, in the melodic send adds them to a queue, and which then uses _send to send queued messages??? It might be a timing issue, or an issue stemming from being async arising from using a queue??? (I am speculating, I might be completely wrong, but I think I have identified the problem.

We need this fixed, because it affects many peoples projects. Please let me know if I can do anything to help fixing it.

Jliporace:
Hi everyone! 
Do we have a recent update or fix on this topic?

doisyg:
Alternatively, can we roll back to an older version without this issue?

altineller:
> Alternatively, can we roll back to an older version without this issue?

replacing SerialClient.py works. not replacing it also works, it just shows the wrong result.

Jliporace:
> > Alternatively, can we roll back to an older version without this issue?
> 
> replacing SerialClient.py works. not replacing it also works, it just shows the wrong result.

@altineller Can, could you link me for what file you replaced SerialClient.py? I tried it by following your previous comment and it didn't work for me. 

And your're referring to replacing the SerialClient.py only in rosserial_python or also replacing the SerialClient.py in rosserial_arduino?

benjinne:
I've tested this solution and it fixes the issue. Why hasn't it been merged in yet?

elXaptation:
Hi, any updates on potential fix for issue #414? Is the recommendation to go down the replacing serialClient.py in the meantime? Thank you.

elXaptation:
>     5\. After hopelessly debugging the problem, in the end, I have replaced the SerialClient.py at my melodic install the SerialClient.py from kinetic install. And it worked! I was able to get service response correctly as True.

Can some one point me in the right direction to "replace SerialClient.py" in melodic with that from kinetic? Is it as simple as it sounds? Thank you.

elXaptation:
Okay. Figured it out... This was my process to "replace SerialClient.py" in Melodic with that from Kinetic.

A few important notes. The Kinetic version of SerialClient.py can be found in the Jade-Devel branch. No Kinetic-Devel Branch that I see. 

(I'm running Ubuntu 18.04)
1. Dive in to python dist-packages

`cd /opt/ros/melodic/lib/python2.7/dist-packages/rosserial_python`

2. Rename existing SerialClient files

```
sudo mv SerialClient.pyc melodicSerialClient.pyc
sudo mv SerialClient.py melodicSerialClient.py
```
4. Browse [SerialClient.py commit before fe1e233 commit](https://github.com/ros-drivers/rosserial/blob/a8c9219eb0aa88dd40bd72ef5bd188fe0bd0f44b/rosserial_python/src/rosserial_python/SerialClient.py)
5. Click "Raw" on right hand side.
6. [select-all], [copy]
7. open new file SerialClient.py. 
```
$ pwd
/opt/ros/melodic/lib/python2.7/dist-packages/rosserial_python
$ sudo vi SerialClient.py
```

8. Paste the copied contents, save
9. Load Arduino IDE ros_lib example "ServiceServer" on your Arduino (Assuming you're using Arduino)
9. Start serial_node:

`rosrun rosserial_python serial_node.py /dev/ttyUSB0`

10. Test service handler:

```
$ rosservice call /test_srv "input: 'true'" 
output: "world"
$ rosservice call /test_srv "input: 'true'" 
output: "hello"
$ rosservice call /test_srv "input: 'true'" 
output: "world"
$ rosservice call /test_srv "input: 'true'" 
output: "hello"
$ rosservice call /test_srv "input: 'true'" 
output: "world"
```
Done. No 
`ERROR: service [/topic] responded with an error: service cannot process request: service handler returned None`

ghost:
@elXaptation 
where you find working ServiceClient.py (https://github.com/ros-drivers/rosserial/releases/tag/0.7.7)
don't forget to update the Arduino ros_lib after step 8
Step 8.1 Delete Arduino/libraries/ros_lib
Step 8.2 Install ros_lib again  (https://wiki.ros.org/rosserial_arduino/Tutorials/Arduino%20IDE%20Setup)
Step 8.3 Make Headers agian (https://wiki.ros.org/rosserial_arduino/Tutorials/Adding%20Custom%20Messages)
Step 8.4 Restart Arduino IDE( If it was Open)
Step 9

gurbain:
Hi!

Is there any reason this issue is still open? I would also really appreciate seeing this fix in the devel branch.

Gabriel

mikepurvis:
The main issue is that I am the sole maintainer at present and my particular use-cases do not involve either `rosserial_python` or the use of services. So without quite a bit of effort, I'm not in a position to evaluate this change— a thread where it works for some people, and others are talking about patching their systems back to the Kinetic version of the file is not super reassuring.

My preferred path would be for someone else to step up (per #515), especially someone who is an actual `rosserial_python` user and willing to take more ownership of it.

Issue #416, opened 2019-02-28
ros-teensy release


sevenbitbyte:
Hi maintainers of ros-drivers. Would you be interested in forking and releasing https://github.com/mcgill-robotics/ros-teensy ?

Its a great package that extends rosserial support into [PJRC's](https://www.pjrc.com/) fantastic teensyduino but has sadly never been released into rosdistro and many of the maintainers seem to have moved on. I think its a good for to be included in ros-drivers or rosserial.

Would be happy to send you a PR if everyone is happy with adding directly to rosserial.

You can follow this issue in ros-teensy -> https://github.com/mcgill-robotics/ros-teensy/issues

PR #417, opened 2019-02-28
Add ros teensy

sevenbitbyte:
# rosserial_teensy

This PR merges the [ros-teensy](https://github.com/mcgill-robotics/ros-teensy) project into rosserial while preserving commit history. This appears to be the best place for this package to be released from as a number of other platforms are supported here.

If this makes sense to all stake holders this seems to be the best path to ensure this project remains well maintained into the future.

## Issues Addressed

* https://github.com/mcgill-robotics/ros-teensy/issues/2
* https://github.com/ros-drivers/rosserial/issues/416

## Future Work

This package does rely upon a blob supplied from PJRC as well as an ARM gcc toolchain. I am not sure how these dependencies may affect some users. There may need to be future work with respect to the blob from the device manufacturer to further streamline usage.


romainreignier:
I don't really get the point of this project.
It is more a CMake config to build a teensyduino project with CMake, there seem to be nothing to do with ROS and rosserial.
I understand that these files could be very handy if someone prefer to use CMake instead of the Arduino "IDE" so they must be maintained somewhere.
But I am not sure a ROS repo is a good place.

sevenbitbyte:
Hi @romainreignier, great question. There are a number of official ROS packages which similarly only provide build support one such example is [ros/cmake_modules](https://github.com/ros/cmake_modules). 

This project, ros-teensy, like wise simply supplies cmake modules though these are primarily useful for consumers of rosserial who wish to use a teensy. Which is why we think this may fit here. Unlike rosserial_aurdino no additional c/c++/ino are needed as the consumer directly imports those dependencies. All other aurdino variants will likely require only cmake files just like the teensy and I'm not sure its ideal user experience to have those artifacts scatter all over the internet.

Here are two projects I've previously used ros-teensy on so you can see how this plays out in practice.

* https://github.com/sevenbitbyte/housebot/tree/master/vega4_head
* https://github.com/sevenbitbyte/housebot/tree/master/vega4_belt

Which drives the LEDs on [this robot](https://www.youtube.com/watch?v=8btrwjQu36c)

romainreignier:
Ok, I see and it make sense to me.

It is indeed quite nice to have  the firmware located in the usual ROS workspace and build it with `catkin_make`.
So packaging these cmake files in a ROS package makes sense and they should even be enhanced in the future to support classic Arduino boards.
Because they are not directly related with the `rossserial` part of `rosserial_arduino`, maybe a new package named something like `arduino_cmake` would make more sense.
But I know that creating, releasing and maintaining a new package is a lot of work that would be reduced by borrowing an existing package.

I will let other users / maintainers commenting on this PR.

Anyway, thanks for the explanation.

doisyg:
> So packaging these cmake files in a ROS package makes sense and they should even be enhanced in the future to support classic Arduino boards.

There is already something for regular Arduino board but it doesn't support the teensy, actually named arduino-cmake ^^:
http://wiki.ros.org/rosserial_arduino/Tutorials/CMake
https://github.com/queezythegreat/arduino-cmake

sevenbitbyte:
@doisyg The core issue here is that ros-teensy has no active maintainers so there is not presently anyone responsible for ensuring future operation of this feature. I've used it in the past and it was a massive help but am not presently using it so I'm also over subscribed to maintain it. As it is effectively a driver, I think the ros-drivers organization is a great place for this to land.

Baycken:
The project itself was made because we didn't fully understand how `arduino-cmake` works. Since `rosserial_arduino` already include `arduino-cmake`, its probably better to bake `teensy-cmake` functionality directly into `arduino-cmake`.

Bei from McGill Robotics

PR #421, opened 2019-03-31
Add getparam handler for serial_node

martonmiklos:
This PR adds ability for the rosserial slaves to access the ros parameters.
The original author of this change is @vmatos . I am just helping to upstream it because he is lack of free time:
https://github.com/vmatos/rosserial/commit/7e3333ebd5e7a3996027b41cb51d6051a350d480#commitcomment-30454859

### DIFF
--- a/rosserial_server/include/rosserial_server/session.h
+++ b/rosserial_server/include/rosserial_server/session.h
@@ -36,6 +36,7 @@
 #define ROSSERIAL_SERVER_SESSION_H
 
 #include <map>
+#include <typeinfo>
 #include <boost/bind.hpp>
 #include <boost/asio.hpp>
 #include <boost/function.hpp>
@@ -44,6 +45,7 @@
 #include <ros/ros.h>
 #include <rosserial_msgs/TopicInfo.h>
 #include <rosserial_msgs/Log.h>
+#include <rosserial_msgs/RequestParam.h>
 #include <topic_tools/shape_shifter.h>
 #include <std_msgs/Time.h>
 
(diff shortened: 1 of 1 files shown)


martonmiklos:
@mikepurvis May I ask for a review on this one? Thanks in advance!

mikepurvis:
In the past, I've resisted adding this functionality without a comprehensive plan for handling retry and so on, but perhaps it's a case where something is better than nothing. Please add a basic test though. Just set a parameter of each type in the .test files (or pull in a common rosparam yaml), and then validate them in the gtest:

https://github.com/ros-drivers/rosserial/tree/melodic-devel/rosserial_test/test
https://github.com/ros-drivers/rosserial/blob/melodic-devel/rosserial_test/src/publish_subscribe.cpp

martonmiklos:
@mikepurvis thanks for the feedback, I will work on getting test coverage for this!

martonmiklos:
I cannot get the getparam working in the test environment. The server side nh's getparam works ok (looks to be exposed from the launch file properly), but the getparam via the client_nh's does not. Any hint would be warmly welcome!

mikepurvis:
The client side one would be the part utilizing your new logic, right? When that request goes through, do you see it in Wireshark?

martonmiklos:
The new logic is on the server side, but the unit test would test it via the client handle. (AFAIK the client side for getparam was implemented before and python server implementation supported it already). 

I see a lot of XMLRPC and TCP traffic as well. Shall I look for it in the XML or it is going to be in the TCP stream (will I need to install the TCPROS dissector?)

PR #425, opened 2019-06-16
Add TX stop function in rosserial_server, by catching SIGINT (ctrl+C)

tongtybj:
I found there was no implementation about the tx stop function while finishing the communication by `ctrl+C` in rosserial_server.
So I added the similar function in [rosserial_python](https://github.com/ros-drivers/rosserial/blob/melodic-devel/rosserial_python/src/rosserial_python/SerialClient.py#L414-L418).

I first tested by adding following few lines:
https://github.com/tongtybj/rosserial/blob/acf73b7627f6910514165415dd6792f4727e0e20/rosserial_server/include/rosserial_server/session.h#L128-L130

But I found the function `stop` was called a few seconds after `ctrl+C`. During the time, the device would still publish topics. So I added the signal handler.

### DIFF
--- a/rosserial_server/include/rosserial_server/session.h
+++ b/rosserial_server/include/rosserial_server/session.h
@@ -50,6 +50,11 @@
 #include "rosserial_server/async_read_buffer.h"
 #include "rosserial_server/topic_handlers.h"
 
+namespace
+{
+  bool terminate_start_flag_ = false;
+};
+
 namespace rosserial_server
 {
 
@@ -85,6 +90,9 @@ public:
     ros_spin_timer_.expires_from_now(ros_spin_interval_);
     ros_spin_timer_.async_wait(boost::bind(&Session::ros_spin_timeout, this,
                                            boost::asio::placeholders::error));
+
+
+    signal(SIGINT, &Session::signal_catch);
   }
 
   Socket& socket()
@@ -116,6 +124,11 @@ public:
 
   void stop()
   {
+    std::cout << "stop!!!" << std::endl;
+    // Send stop tx command to MCU
+    //std::vector<uint8_t> message(0);
+    //write_message(message, rosserial_msgs::TopicInfo::ID_TX_STOP);
+
     // Abort any pending ROS callbacks.
     ros_callback_queue_.clear();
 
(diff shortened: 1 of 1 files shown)


rosserial_server/include/rosserial_server/session.h
```diff
@@ -116,6 +124,11 @@ class Session : boost::noncopyable
 
   void stop()
   {
+    std::cout << "stop!!!" << std::endl;
+    // Send stop tx command to MCU
+    //std::vector<uint8_t> message(0);
+    //write_message(message, rosserial_msgs::TopicInfo::ID_TX_STOP);
```
mikepurvis: Please clean this up, and replace the `cout` with `ROS_INFO_STREAM` or `ROS_DEBUG_STREAM` as appropriate.

rosserial_server/include/rosserial_server/session.h
```diff
@@ -272,6 +295,7 @@ class Session : boost::noncopyable
     ros::serialization::OStream stream(&buffer_ptr->at(0), buffer_ptr->size());
     uint8_t msg_len_checksum = 255 - checksum(message.size());
     stream << (uint16_t)0xfeff << (uint16_t)message.size() << msg_len_checksum << topic_id;
+    std::cout << "message size: " << message.size() << ", msg_len_checksum:" << topic_id << std::endl;
```
mikepurvis: `ROS_DEBUG_STREAM` here, please.

rosserial_server/include/rosserial_server/session.h
```diff
@@ -383,6 +407,11 @@ class Session : boost::noncopyable
     return (val >> 8) + val;
   }
 
+  static void signal_catch(int sig)
+  {
+    terminate_start_flag_ = true;
```
mikepurvis, line 409: If we're catching this, do we need to manually propagate it? Per http://wiki.ros.org/roscpp/Overview/Initialization%20and%20Shutdown, I think we need a `ros::shutdown();` here.

mikepurvis (changes requested):

rosserial_server/include/rosserial_server/session.h
tongtybj: Do you mean to move the following action inside this static function?

https://github.com/ros-drivers/rosserial/blob/acf73b7627f6910514165415dd6792f4727e0e20/rosserial_server/include/rosserial_server/session.h#L178-L181

Or do you mean that we do not need to send `rosserial_msgs::TopicInfo::ID_TX_STOP`, and only need to do `ros:shutdown()` for the stop action?

rosserial_server/include/rosserial_server/session.h
tongtybj: Revised in next commit.

rosserial_server/include/rosserial_server/session.h
tongtybj: Revised in next commit.

PR #426, opened 2019-06-18
Change the access modifiers "private" to "protected" in NodeHandle_

tongtybj:
Several member functions in `NodeHandle_` can be overwritten (e.g. `spinOnce`), so this class was designed to afford the potential derivation usage. However, it is impossible to access to the private variables prevents from the derived class (This is the problem what I am facing with). So I recommend to change the variables to be `protected `.


mikepurvis:
Seems reasonable to me, thanks!

PR #432, opened 2019-07-08
Add rosservice server function in rosserial_server.

tongtybj:
So far, the rosservice server function is not implemented in rosserial_server,  since c/c++ is static typing.
Thus, it is difficult to do like python: https://github.com/ros-drivers/rosserial/blob/melodic-devel/rosserial_python/src/rosserial_python/SerialClient.py#L153-L159

However, it is quite inconvenient  that there is no rosservice server in c/c++ mode. 

This PR is aimed to achieve the rosservice server by adding an additional python script:
`rosserial_server/rosservice_server_manager.py`, which is  the  extracted part from `rosserial_python/SerialClient.py`. This script communicates with `rosserial_server/session.h` by a new message: `SerializedMessage.msg`. 

I have testified with arduino sample: `ServiceServer`.

I know the duplication of the existing code is not clever. Maybe it is better to derive from `rosserial_python/SerialClient.py`.

### DIFF
--- a/rosserial_server/CMakeLists.txt
+++ b/rosserial_server/CMakeLists.txt
@@ -6,12 +6,25 @@ find_package(catkin REQUIRED COMPONENTS
   rosserial_msgs
   std_msgs
   topic_tools
+  message_generation
 )
 
 find_package(Boost REQUIRED COMPONENTS
   system
   thread
 )
+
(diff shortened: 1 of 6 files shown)


tongtybj:
refactoring by #453

PR #442, opened 2019-09-29
AVR Float64 conversion: fix special cases like denormal numbers & inf

PetteriAimonen:
Currently the AVR 64-bit float conversion doesn't correctly handle denormal numbers, NaN or infinity values. For example `nan` becomes `inf` and `9.9999994e-39` becomes `0`.

I know they're not very big errors for practical purposes. I just happened to have figured this out in the past so in case you are interested in a fix here it is.


mikepurvis:
Sorry for the late attention here, but thank you for this contribution.

PR #446, opened 2019-11-05
Drop separate node for message service

mikepurvis:
An effort to eliminate the separate and trivial "message info" node from `rosserial_server`, by embedding the Python interpreter itself and accessing message md5sum and full_text attributes that way.

Motivation is that we were having issues with 5 seconds not being a long enough wait for the service to become available, resulting in bags without the full_text attribute stored (which `rosbag` then complains about).

FYI @dlguo-cpr


zeroMG (commented):
LGTM

dlguo-cpr:
LGTM

PR #448, opened 2019-11-13
Add serial communication for ESP controllers 

PouceHeure:
Hello, 

## reason: 

Due to the clock frequency, I decided to use an ESP, but for personal reason I prefer to use it by USB serial. 

May be, this option can help someone. 

## use: 
- define ESP_SERIAL: use the esp in "serial mode"
- not define ESP_SERIAL: use the esp in "tcp mode"

## test: 
- ros version: melodic 
- boards tested: ESP32 and ESP8266

### DIFF
--- a/rosserial_arduino/src/ros_lib/ros.h
+++ b/rosserial_arduino/src/ros_lib/ros.h
@@ -37,7 +37,7 @@
 
 #include "ros/node_handle.h"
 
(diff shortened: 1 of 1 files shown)


romainreignier:
This looks ok for me.

mikepurvis:
Looks reasonable.

CorentinChauvin:
Hello,

I find myself in the same situation, where I'd like to use an ESP through serial, with ROS Noetic. Would it make sense to do the same changes on the noetic branch?

Regards,
Corentin

PR #449, opened 2019-11-18
Fix bug: assign the md5 for service

tongtybj:
Fix the bug related to #446 .

`serivce_md5_` is used in following function:
https://github.com/tongtybj/rosserial/blob/9f9ab3ee71e00e48c69032096ec9e96f896a9724/rosserial_server/include/rosserial_server/topic_handlers.h#L173


mikepurvis:
Perfect, thanks for checking this!

PR #453, opened 2019-11-21
[WIP] Implement rosservice server in rosserial_server

tongtybj:
Use `topic_tools::ShapeShifter` to create  `proxy` service servers in rosserial server.

Also, during the implementation, I found a very interesting phenomenon.
When the callback function of a `proxy` service server is called,  all process related to `boost::asio` seem to be paused, meaning that no response can be caught from rosserial client unless this service callback function is over. This can be check in c32952f .

So I replace `boost::asio::deadline_timer ros_spin_timer_` with `ros::AsyncSpinner` which has 2 threads. I know this might be a trivial reverting action. But can anyone tell me the motivation to use `boost::asio::deadline_timer ros_spin_timer_` and what [this comment](https://github.com/ros-drivers/rosserial/blob/melodic-devel/rosserial_server/include/rosserial_server/session.h#L156-L157)  means?


romainreignier:
Hi, nice PR!
I have never needed a service server on the `rosserial_client` side but it is nice to see this feature added.

About the async spinner, from what I understand, the current implementation is using the io_service to keep only one thread to simplify the design and avoid synchronization issues.
In your implementation, there is now 3 threads and I do not see any synchronization primitives to protect the resources.

rosserial_server/include/rosserial_server/topic_handlers.h
```diff
@@ -184,17 +184,117 @@ class ServiceClient {
...
+        ros::Duration(0.01).sleep();
+
+        if (cnt++ > 500) // hard-coding
+          {
+            ROS_WARN_STREAM(service_server_.getService() << ": no response!!");
+            return false;
+          }
+      }
+
+    response_message = response_message_;
```
rezeck: Hi, I am getting segfault at this line after I call a custom service at the third time. Any tip how to fix it? I comment this code line and the segfault disappears, but consequently, I do not receive the response message.

tongtybj: @rezeck

Sorry for the very late response. 
~~I have tested  with https://github.com/ros-drivers/rosserial/blob/noetic-devel/rosserial_arduino/src/ros_lib/examples/ServiceServer/ServiceServer.pde, however I can not encounter the problem you pointed out. Could you paste your code to let me test.~~

~~By the way, as @romainreignier pointed out,  there is no  synchronization primitives to protect the resources in this commit, which might be the reason of  your problem. I will add mutex in multiple threads.~~

tongtybj: I have carefully studied this problem again, found there is a fatal bug due to `response_message = response_message_;`
`response_message` which the type is [topic_tools::ShapeShifter](http://docs.ros.org/kinetic/api/topic_tools/html/classtopic__tools_1_1ShapeShifter.html), has a pointer type member `uint8_t * msgBuf`. 

 It seems like the `response_message` will be deleted after this rosservice server callback function, which leads to a invalid access to `msgBuf` in   `response_message_`, in next rosservice response.

I have fixed this problem by commit  3ffc8e4 . Could you please help me to check the latest version of the branch?

tongtybj:
@romainreignier 

> In your implementation, there is now 3 threads and I do not see any synchronization primitives to protect the resources.

I think the only shared resource that should be protected is the write (`async_write`) process, since the ros callback functions and several dealine_timers do the write process from different threads.
 
 So I add a mutex to  lock the following part:

https://github.com/ros-drivers/rosserial/blob/50a94094e334e7c99d41e6138a5bb5dcdd02b7ab/rosserial_server/include/rosserial_server/session.h#L267-L270

PR #454, opened 2019-11-23
Fix makedirs error

fdila:
If you previously generated the libraries in your project you couldn't regenerate them because this script would try to create an existing directory.
Now it checks if the directory is already present and, in that case, it deletes the folder before recreating it


romainreignier:
This fix is nice, but I am not sure it is sufficient.
Usually, this `make_library` script is called from a `make_libraries.py` in an implementation of `rosserial_client` like `rosserial_arduino` and this script will first fail if the directory `ros_lib` already exists.

PR #457, opened 2020-01-08
rosserial_server: update install rules for binary targets

meyerj:
- Split install rule for binary targets in one for the library and for the executables (required for Windows DLLs?)
- `ARCHIVE DESTINATION` was missing in the install rule for binary targets, which prevented building and installing `rosserial` as static binaries.

See also http://wiki.ros.org/catkin/CMakeLists.txt#Optional_Step:_Specifying_Installable_Targets.

### DIFF
--- a/rosserial_server/CMakeLists.txt
+++ b/rosserial_server/CMakeLists.txt
@@ -50,15 +50,20 @@ target_link_libraries(${PROJECT_NAME}_udp_socket_node ${catkin_LIBRARIES} ${PROJ
 set_target_properties(${PROJECT_NAME}_udp_socket_node PROPERTIES OUTPUT_NAME udp_socket_node PREFIX "")
 add_dependencies(${PROJECT_NAME}_udp_socket_node ${catkin_EXPORTED_TARGETS})
 
-
 install(
   TARGETS
     ${PROJECT_NAME}_lookup
+  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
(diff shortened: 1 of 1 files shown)


mikepurvis:
If you add `ARCHIVE DESTINATION`, is it still necessary to split the `install` rule? My impression has been that when specifying multiple destinations, CMake evaluates which is most appropriate on a target by target basis.

meyerj:
> If you add `ARCHIVE DESTINATION`, is it still necessary to split the `install` rule? My impression has been that when specifying multiple destinations, CMake evaluates which is most appropriate on a target by target basis.

Yes, for static libraries and executables on Linux adding `ARCHIVE DESTINATION` would be enough. But Windows requires that shared libraries (dlls) are installed to `${CATKIN_GLOBAL_BIN_DESTINATION}`, while executable targets should be installed to `${CATKIN_PACKAGE_BIN_DESTINATION}`. That is why the two install rules need to be split, because both are `RUNTIME` targets.

Reference:
- http://wiki.ros.org/action/diff/catkin/CMakeLists.txt?action=diff&rev1=83&rev2=84
- http://docs.ros.org/melodic/api/catkin/html/howto/format1/building_libraries.html

Not sure whether rosserial support Windows at all and I am not a Windows user. It was more to follow the latest guidelines. The main intention of the PR was to add the `ARCHIVE DESTINATION` or the library target.

mikepurvis:
There is no intention of Windows support, but I don't think anything particularly precludes it either. Thanks for the full rationale.

PR #460, opened 2020-01-16
Added target ID to support Teensy 4.0

aaraujo11:


mikepurvis:
Thanks for this contribution.

PR #463, opened 2020-02-06
[Windows] use c++ signed trait to replace ssize_t for better portability.

seanyen:
The `ssize_t` is a type defined in POSIX standard. However, on non-POSIX system (for example, Windows), this could be absent.

This change is inspired by an alternative being talked about in this conclusion: [`LWG issue: 2251. C++ library should define ssize_t`](https://cplusplus.github.io/LWG/issue2251), which mentions that this can be achieved by C++11 type trait feature.

### DIFF
--- a/rosserial_server/include/rosserial_server/async_read_buffer.h
+++ b/rosserial_server/include/rosserial_server/async_read_buffer.h
@@ -40,6 +40,10 @@
 
 #include <ros/ros.h>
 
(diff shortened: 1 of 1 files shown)


seanyen:
@mikepurvis  This is ready for review and merge. Thanks!

mikepurvis:
Great, thanks!

PR #465, opened 2020-02-16
Update SerialClient.py

zaizhizhuang:
The 'send' function need return a vaule .
        ServiceServer
            callback
                if self.parent.send(self.id, data_buffer.getvalue()) >= 0:

### DIFF
--- a/rosserial_python/src/rosserial_python/SerialClient.py
+++ b/rosserial_python/src/rosserial_python/SerialClient.py
@@ -740,6 +740,13 @@ class SerialClient(object):
         Queues data to be written to the serial port.
         """
         self.write_queue.put((topic, msg))
+        """
+        Here need return a vaule .zzz
+        ServiceServer
+            callback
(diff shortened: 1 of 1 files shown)


rosserial_python/src/rosserial_python/SerialClient.py
```diff
@@ -740,6 +740,13 @@ def send(self, topic, msg):
         Queues data to be written to the serial port.
         """
         self.write_queue.put((topic, msg))
+        """
+        Here need return a vaule .zzz
+        ServiceServer
+            callback
+                if self.parent.send(self.id, data_buffer.getvalue()) >= 0:
+        """
+        return 1
```
mikepurvis, line 749: The queue is not limited in length, so there is no way for this to fail. IMO the issue should be fixed in ServiceServer.

mikepurvis (changes requested):

PR #469, opened 2020-03-03
fix py3 print compat and trailing whitespaces

acxz:


mikepurvis:
Great, thanks! I'm  working on a larger attempted port of rosserial_python to py3 but it's a bit of a slog. This is a good start. :)

PR #470, opened 2020-03-04
make except call py3 compat

acxz:
this one flew by me, should be good now


acxz:
@mikepurvis can u take a look at this one rq?

mikepurvis:
Merged for now, but there's quite a bit more that needs to be updated than just this one thing.

acxz:
For sure.

PR #473, opened 2020-03-18
rosserial_python: Catch exception with 'as' instead of comma

stertingen:
See https://stackoverflow.com/questions/2535760/python-try-except-comma-vs-as-in-except.

Python 3 enforces the usage of `as`, so fix it here.


stertingen:
Duplicate of https://github.com/ros-drivers/rosserial/pull/470

Issue #277, opened 2017-03-09
Sync lost when publishing Odometry message. 


AlexisTM:
Dear, 

* Hardware : `Mega2560`
* Software : `Indigo, 0.6.4`

I keep having Sync issues when I am trying to send the nav_msgs/Odometry message. 

I have multiple other topics, LEDs subscriber to a std_msgs/ColorRGBA topic, US sensors (non blocking architecture with NewPing library) publishing a std_msgs/Bool topic. All those are working if I do not send the Odometry message (commenting off  `rover_odom.publish(&odometry_msg);`) but if I uncomment it, simply loose Sync all the time. 

Is the message too big?

```
boolean odometrySender(){
    analogWrite(LED_R_PIN, 255);
    analogWrite(LED_G_PIN, 0);
    analogWrite(LED_B_PIN, 0);
    now_encoder_time = millis()/1000.0;
    if(!updated_left) return false;
    if(!updated_right) return false;

    delta_left = current_encoder_left - last_encoder_left;
    delta_right = current_encoder_right - last_encoder_right;

    omega_left = (delta_left * distance_per_tick) / (now_encoder_time - last_encoder_time);
    omega_right = (delta_right * distance_per_tick) / (now_encoder_time - last_encoder_time);

    v_left = omega_left * WHEEL_RADIUS; //radius
    v_right = omega_right * WHEEL_RADIUS;

    vx = ((v_right + v_left) / 2)*10; // Times 10 ?
    vy = 0;
    vth = ((v_right - v_left)/distance_between_wheels)*10;

    double dt = (now_encoder_time - last_encoder_time);
    double delta_x = (vx * cos(th)) * dt;
    double delta_y = (vx * sin(th)) * dt;
    double delta_th = vth * dt;

    x += delta_x;
    y += delta_y;
    th += delta_th;

    //geometry_msgs::Quaternion odom_quat = tf::createQuaternionMsgFromYaw(th);
/*
    geometry_msgs::TransformStamped odom_trans;
    odom_trans.header.stamp = ros_now_encoder_time;
    odom_trans.header.frame_id = "odom";
    odom_trans.child_frame_id = "base_link";

    odom_trans.transform.translation.x = x;
    odom_trans.transform.translation.y = y;
    odom_trans.transform.translation.z = 0.0;
    //odom_trans.transform.rotation = odom_quat;

    //send the transform
    odom_broadcaster.sendTransform(odom_trans);
*/
    //Odometry message

[...]

Issue #415, opened 2019-02-27
Arduino Due _SAM3XA_ baud_rate limitation at 115200


lancemk:
My project require to publish IMU topic up to **150Hz** using **Arduino DUE** using rosserial_python node_serial.py. We hooked up DUE and MPU9250 via i2c connection, and DUE transfer data stream via USB 2.0. Arduino sketch is as shown below:
```
#define _SAM3XA_
#include <ros.h>
#include <tf/transform_broadcaster.h>
#include <geometry_msgs/Vector3.h>
#include <sensor_msgs/Imu.h>

ros::NodeHandle_<ArduinoHardware, 5, 5, 256, 512> nh;   
sensor_msgs::Imu imu_msg;
ros::Publisher imu_topic("IMU_TOPIC", &imu_msg);

char base_link[] = "/base_link";
char odom[] = "/odom";

void setup() {
  nh.getHardware()->setBaud(115200);
  nh.initNode();
  nh.advertise(imu_topic);

  // start communication with IMU 
  status = IMU.begin();
  if (status < 0) {
    Serial.println("IMU initialization unsuccessful");
    Serial.println("Check IMU wiring or try cycling power");
    Serial.print("Status: ");
    Serial.println(status);
    while(1) {}
  }
  nh.loginfo("IMU initialization successful");
 // setting the IMU
}

void loop() {
   
  IMU.readSensor();
  if (nh.connected()) {
  imu_msg.header.stamp = nh.now();
  imu_msg.header.frame_id = odom;
  
  imu_msg.linear_acceleration.x = // <TYPE_IMU_ACCEL> ;
  imu_msg.linear_acceleration.y = // <TYPE_IMU_ACCEL>;
  imu_msg.linear_acceleration.z = // <TYPE_IMU_ACCEL>;

  imu_msg.angular_velocity.x = // <TYPE_IMU_GYRO>; 
  imu_msg.angular_velocity.y = // <TYPE_IMU_GYRO>; 
  imu_msg.angular_velocity.z = // <TYPE_IMU_GYRO>; 

  imu_topic.publish(&imu_msg);

  } else {
    if(false) Serial.println("Not Connected");
  }
  nh.spinOnce();
  delayMicroseconds(1250);
}
```

Current topic frequency of maximum **36Hz** by running `rostopic hz /<TOPIC_NAME>` at **baud_rate = 115200**. A higher baud rate setting node_serial node shows error:

[...]

Issue #181, opened 2015-04-03
Performance issue of rosserial_arduino


ymkim92:
Hi,

I have used rosserial_arduino to connect Arduino to my roscore on Linux.
When I run roscore on my laptop. The performance is quite stable.
I am sending data at 100Hz rate in both direction.
But, on the embedded Linux of Gumstix overo, its performance is very unstable.
It is changing in range of 80Hz ~ 100Hz and sometimes the connection is broken with this error:

```
[ERROR] [WallTime: 1428040762.537858] Lost sync with device, restarting...
```

Can I fix this problem?

mikepurvis:
I don't have a lot of experience trying to run the ROS side on resource-constrained platforms, but I would suggest trying [rosserial_server](http://wiki.ros.org/rosserial_server#Usage) in place of rosserial_python. It's a C++ implementation of the server, and more performance-oriented.

ymkim92:
Thanks a lot. It is working well in my laptop. Then, I tried to run rosserial_server on my Gumstix. But, they don't support rosserial_server. I think it is still unstable and in development.

mikepurvis:
Hmm, is it just that rosserial_server isn't available in the armhf debians? If you tried to compile it yourself and ran into an issue, I'd be interested to have that reported so it can be addressed.

ymkim92:
I compiled rosserial_server and install it in Gumstix. The performance is much more stable. Great!
It is almost the same as on my laptop with rosserial_python. But still I have a problem. That is both ROS and Arduino did not make stable timing in 100Hz.

PR #475, opened 2020-03-24
Use arduino packet write function.

rikba:
Users have observed that they cannot publish large ROS messages on an Arduino at a high rate https://github.com/ros-drivers/rosserial/issues/277. Their workaround was to create custom messages with minimized size.

We timed the [publish function](https://github.com/ros-drivers/rosserial/blob/melodic-devel/rosserial_client/src/ros_lib/ros/node_handle.h#L511-L545) and realized that [serial writing](https://github.com/ros-drivers/rosserial/blob/melodic-devel/rosserial_client/src/ros_lib/ros/node_handle.h#L537) takes the major share.

Publishing messages in packets rather than character by character [speeds up serial transmission](https://forum.arduino.cc/index.php?topic=575333.0) significantly. 

Here we publish IMU messages (320 bytes) on a SAMD21 microcontroller through native USB and time the serial write function.

Sending bytewise about 9000us per message (~284 kb/s) (**old**)
serial_write_broken

Sending packetwise about 450us per message (~5.689 Mb/s) (**new**)
serial_write_fixed

May also be relevant to @floriantschopp

### DIFF
--- a/rosserial_arduino/src/ros_lib/ArduinoHardware.h
+++ b/rosserial_arduino/src/ros_lib/ArduinoHardware.h
@@ -101,8 +101,7 @@ class ArduinoHardware {
 
     int read(){return iostream->read();};
     void write(uint8_t* data, int length){
-      for(int i=0; i<length; i++)
-        iostream->write(data[i]);
+      iostream->write(data, length);
     }
 
     unsigned long time(){return millis();}


rikba:
This may also help with https://github.com/ros-drivers/rosserial/issues/181, https://github.com/ros-drivers/rosserial/issues/415 where the users observe dropped message rates.

romainreignier:
This change looks like a nice addition.
Did you test it on low end 8-bit Arduino with real serial port and USB to serial converter (like UNO)?

rikba:
I did only test it on a native USB device. I have an Arduino Micro lying around where I could implement a sketch and time the message throughput.

romainreignier:
But Arduino Micro is still Native USB.
Anyway, I don't think this change could hurt.

mikepurvis:
Looks like a great change, thanks!

PR #476, opened 2020-04-10
Call Py_Finalize before throwing exception

stertingen:


mikepurvis:
Good catch, thank you. Potentially a better way would be to wrap the whole thing in a try-catch and then re-raise the exception and perform the finalize afterward. But the logic is simple enough that I think this is probably fine for now.

PR #479, opened 2020-04-13
Make deprecation message a warning and more specific

stertingen:
See: https://wiki.ros.org/Verbosity%20Levels


mikepurvis:
Thanks, will merge this to `melodic-devel` and the removal will happen in Noetic.

PR #480, opened 2020-04-14
Fixes for new warnings/behaviours in GCC7+/cppcheck.

mikepurvis:
A bunch of stuff that came up when testing with the GCC9-based toolchain that will ship in Ubuntu Focal:
- Add missing member initializations.
- Add missing override specifiers.
- Hacky cast for the 16 bit AVR float logic.

There are a few more member initializations to add in NodeHandle. Will get to those before merging.

Also, I believe this will break Bionic, and it's a pretty breaking change regardless, so this probably merits cutting a `noetic-devel` branch for it.

### DIFF
--- a/rosserial_client/src/ros_lib/ros/msg.h
+++ b/rosserial_client/src/ros_lib/ros/msg.h
@@ -53,7 +53,7 @@ public:
   /**
    * @brief This tricky function handles promoting a 32bit float to a 64bit
    *        double, so that AVR can publish messages containing float64
-   *        fields, despite AVV having no native support for double.
+   *        fields, despite AVR having no native support for double.
    *
    * @param[out] outbuffer pointer for buffer to serialize to.
    * @param[in] f value to serialize.
@@ -63,7 +63,11 @@ public:
    */
   static int serializeAvrFloat64(unsigned char* outbuffer, const float f)
   {
-    const int32_t* val = (int32_t*) &f;
+    // Cast through void* to silence a GCC warning about this being a
+    // non-portable cast (we don't care, it only has to work on AVR).
+    const void* vval = reinterpret_cast<const void*>(&f);
+    const int32_t* val = reinterpret_cast<const int32_t*>(vval);
+
     int32_t exp = ((*val >> 23) & 255);
     if (exp != 0)
     {
@@ -101,7 +105,10 @@ public:
    */
   static int deserializeAvrFloat64(const unsigned char* inbuffer, float* f)
   {
-    uint32_t* val = (uint32_t*)f;
+    // Cast through void* to silence a GCC warning about this being a
+    // non-portable cast (we don't care, it only has to work on AVR).
+    void* vval = reinterpret_cast<void*>(f);
+    uint32_t* val = reinterpret_cast<uint32_t*>(vval);
     inbuffer += 3;
 
     // Copy truncated mantissa.
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -104,19 +104,19 @@ template<class Hardware,
 class NodeHandle_ : public NodeHandleBase_
 {
 protected:
-  Hardware hardware_;
+  Hardware hardware_{};
 
   /* time used for syncing */
-  uint32_t rt_time;
+  uint32_t rt_time{0};
 
   /* used for computing current time */
-  uint32_t sec_offset, nsec_offset;
+  uint32_t sec_offset{0}, nsec_offset{0};
 
   /* Spinonce maximum work timeout */
-  uint32_t spin_timeout_;
+  uint32_t spin_timeout_{0};
 
-  uint8_t message_in[INPUT_SIZE];
-  uint8_t message_out[OUTPUT_SIZE];
+  uint8_t message_in[INPUT_SIZE] = {0};
+  uint8_t message_out[OUTPUT_SIZE] = {0};
 
   Publisher * publishers[MAX_PUBLISHERS];
   Subscriber_ * subscribers[MAX_SUBSCRIBERS];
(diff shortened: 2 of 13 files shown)


rosserial_client/src/ros_lib/ros/msg.h
```diff
@@ -63,7 +63,11 @@ class Msg
    */
   static int serializeAvrFloat64(unsigned char* outbuffer, const float f)
   {
-    const int32_t* val = (int32_t*) &f;
+    // Cast through void* to silence a GCC warning about this being a
+    // non-portable cast (we don't care, it only has to work on AVR).
+    const void* vval = reinterpret_cast<const void*>(&f);
+    const int32_t* val = reinterpret_cast<const int32_t*>(vval);
```
mikepurvis, line 69: Could potentially slip this by the compiler using a `union` as well, but this is nice and explicit. We're doing a bad thing, and we should feel bad.

romainreignier:
About improving code quality, it would be nice to make the time related functions `const` like I did in our [qt version](https://github.com/1r0b1n0/rosserial_qt/commit/382664c228ca5a7615d129801359ca15ad5e781f).

Also, we could align the current [`toNsec`](https://github.com/ros-drivers/rosserial/blob/6bf5794c9169f9c514f1cb52c0a4f7cf86047744/rosserial_client/src/ros_lib/ros/time.h#L67) with the ROS [`toNSec`](https://github.com/ros/roscpp_core/blob/b2a9c0affcaacb510b6e9ec6e46b595bb145d3c9/rostime/include/ros/time.h#L156) (capital `S`).
This is breaking the API but if a new `noetic-devel` branch is created, maybe it's the moment.

Regarding time functions, maybe we could try to use 64 bit variables even if I know that the major use case is on small microcontrollers so the time is not always an epoch time.
What do you think about that?

mikepurvis:
PR is up for noetic-devel branch PR testing here: https://github.com/ros/rosdistro/pull/24473

The tests will continue to fail until the Python 3 situation in `rosserial_python` is resolved. I might simply disable those tests in this PR so that it can merge and then open a separate one to get `rosserial_python` functional again.

Note also that the travis build will also have to be updated; there's no base VM for focal, so it should probably be switched to use the docker service, see: https://docs.travis-ci.com/user/docker/ (though that would be blocked on having the official ROS Noetic docker image— https://github.com/osrf/docker_images/pull/395)

rosserial_python/src/rosserial_python/SerialClient.py
```diff
@@ -545,7 +538,7 @@ def run(self):
                 else:
                     rospy.loginfo("wrong checksum for topic id and msg")
 
-            except IOError as exc:
+            except ImportError as exc:
```
stertingen: What could possibly raise an `ImportError` that is caught here?

rosserial_python/src/rosserial_python/SerialClient.py
```diff
@@ -311,23 +312,28 @@ def inWaiting(self):
             if chunk == '':
```
stertingen: This should be `if chunk == b'':` I think.

rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -104,19 +104,19 @@ template<class Hardware,
...
-  uint32_t spin_timeout_;
+  uint32_t spin_timeout_{0};
 
-  uint8_t message_in[INPUT_SIZE];
-  uint8_t message_out[OUTPUT_SIZE];
+  uint8_t message_in[INPUT_SIZE] = {0};
+  uint8_t message_out[OUTPUT_SIZE] = {0};
 
   Publisher * publishers[MAX_PUBLISHERS];
   Subscriber_ * subscribers[MAX_SUBSCRIBERS];
```
stertingen: For the sake of consistency, `publishers` and `subscribers` should be initialized with ` = {nullptr}`.

rosserial_python/src/rosserial_python/SerialClient.py
```diff
@@ -288,13 +286,16 @@ def write(self, data):
         totalsent = 0
 
         while totalsent < length:
-            sent = self.socket.send(data[totalsent:])
+            try:
+                sent = self.socket.send(data[totalsent:])
+            except BrokenPipeError:
+                sent = 0
```
stertingen: Why not `raise RuntimeError(...)` here?

The message "RosSerialServer.write() socket connection broken" below seems inappopriate for the case in which a `BrokenPipeError` was not caught, but zero bytes were sent anyway.

stertingen (commented):
Although I approve Python 3 support in general; I wonder why is it included in this PR (targeting C++warnings).

mikepurvis:
@stertingen Without proper Python 3 support landing for Noetic, the tests using `rosserial_python` don't pass. However, given the scope of the changes, perhaps it would make more sense to just disable those, merge the smaller original change here, and make a separate PR for fixing `rosserial_python`.

rosserial_python/src/rosserial_python/SerialClient.py
mikepurvis: I agree, this makes no sense. Reverting.

rosserial_client/src/ros_lib/ros/node_handle.h
mikepurvis: Done.

rosserial_python/src/rosserial_python/SerialClient.py
mikepurvis: Agree, changed.

rosserial_client/src/ros_lib/ros/node_handle.h
mikepurvis: Fixed. I wonder why cppcheck complained about all the other ones and not these.

rosserial_python/src/rosserial_python/SerialClient.py
mikepurvis: Hmm, yeah I wasn't sure what to do here. From https://stackoverflow.com/a/34920917/109517, it seems like `socket.send` can legitimately return zero in at least some non-error scenarios. I suppose the risk is that it would hard loop on this, especially if you had a non-blocking socket, but that's not the case here with a dedicated writer thread. Have updated it to ignore the 0 scenario and instead throw from inside the `except`.

rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -104,52 +104,27 @@ template<class Hardware,
...
 
-  uint8_t message_in[INPUT_SIZE];
-  uint8_t message_out[OUTPUT_SIZE];
+  uint8_t message_in[INPUT_SIZE] = {0};
+  uint8_t message_out[OUTPUT_SIZE] = {0};
 
-  Publisher * publishers[MAX_PUBLISHERS];
-  Subscriber_ * subscribers[MAX_SUBSCRIBERS];
+  Publisher * publishers[MAX_PUBLISHERS] = {nullptr};
+  Subscriber_ * subscribers[MAX_SUBSCRIBERS] {nullptr};
```
stertingen, line 122:
```suggestion
  Subscriber_ * subscribers[MAX_SUBSCRIBERS] = {nullptr};
```

mikepurvis:
Closed in favour of #508.

asukiaaa:
Thank you for the information.
I merged your commit to my branch [asukiaaa:feature/noetic-python-support](https://github.com/asukiaaa/rosserial/tree/feature/noetic-python-support) to activate test.
Could you create PR to my(asukiaaa's) branch not to mikepurvis's branch on next?

PR #481, opened 2020-04-14
Removed unused service client for message info service

stertingen:


mikepurvis:
Ha, good catch, thanks.

PR #482, opened 2020-04-14
Fix warning when using std_msgs/Empty

stertingen:
See also https://github.com/ros/ros_comm/issues/344 and according fix

### DIFF
--- a/rosserial_server/src/msg_lookup.cpp
+++ b/rosserial_server/src/msg_lookup.cpp
@@ -73,14 +73,27 @@ const MsgInfo lookupMessage(const std::string& message_type, const std::string s
   Py_XDECREF(msg_class);
 
 #if PY_VERSION_HEX > 0x03000000
-  full_text = full_text ? full_text : PyUnicode_New(0, 0);
-  msginfo.full_text.assign(PyUnicode_AsUTF8(full_text));
+  if (full_text)
+  {
+    msginfo.full_text.assign(PyUnicode_AsUTF8(full_text));
+  }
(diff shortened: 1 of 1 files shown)


rosserial_server/src/msg_lookup.cpp
```diff
@@ -73,14 +73,27 @@ const MsgInfo lookupMessage(const std::string& message_type, const std::string s
...
+    msginfo.full_text.assign(PyUnicode_AsUTF8(full_text));
+  }
   msginfo.md5sum.assign(PyUnicode_AsUTF8(md5sum));
 #else
-  full_text = full_text ? full_text : PyString_FromString("");
-  msginfo.full_text.assign(PyString_AsString(full_text));
+  if (full_text)
+  {
+    msginfo.full_text.assign(PyString_AsString(full_text));
+  }
```
mikepurvis, line 85: What is the purpose of this part of the change?

stertingen: `full_text` might be a `nullptr` at this point, that's why there was the assignment of empty pthon string (`PyString_FromString("")` or `PyUnicode_New(0, 0)`) to `full_text`; only to be converted into a `std::string` later.

`msg.full_text` is an empty string by default, so we can skip the whole creation of an empty python string and just assign the python string if it is not a `nullptr`.

For the next part, it is important for `full_text` to stay a `nullptr` if it was before, so we explicitly do not insert the empty line when the lookup somehow failed.

mikepurvis: Right, that makes sense. Thanks for the fix.

PR #483, opened 2020-04-14
ros_lib: Removed code duplication in node_handle.h

stertingen:
Also, do not generate multiple template instances of
NodeHandle::subscribe()

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -428,14 +428,13 @@ public:
   }
 
   /* Register a new subscriber */
-  template<typename SubscriberT>
-  bool subscribe(SubscriberT& s)
+  bool subscribe(Subscriber_& s)
   {
     for (int i = 0; i < MAX_SUBSCRIBERS; i++)
     {
       if (subscribers[i] == 0) // empty slot
       {
-        subscribers[i] = static_cast<Subscriber_*>(&s);
+        subscribers[i] = &s;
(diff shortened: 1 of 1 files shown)


rosserial_client/src/ros_lib/ros/node_handle.h
```diff
@@ -448,33 +447,17 @@ class NodeHandle_ : public NodeHandleBase_
   bool advertiseService(ServiceServer<MReq, MRes, ObjT>& srv)
   {
     bool v = advertise(srv.pub);
-    for (int i = 0; i < MAX_SUBSCRIBERS; i++)
-    {
-      if (subscribers[i] == 0) // empty slot
-      {
-        subscribers[i] = static_cast<Subscriber_*>(&srv);
-        srv.id_ = i + 100;
```
mikepurvis, line 456: What is the impact of losing this logic?

stertingen: It's exactly the same code as in `subscribe()`. Also, the return value has not changed; `advertiseService()` returns only `true` if both publisher and subscriber has been created; furthermore the subscriber is created regardless of the value of `v`.

Because `advertiseService` and `serviceClient` now only consist of two function calls (and the `&&` if we keep the return value), these functions might be more likely inlined by the compiler.

mikepurvis: Okay, yes, this makes sense. Thanks for explaining.

mikepurvis (approved):

PR #485, opened 2020-04-15
WIP: Add handler for parameter requests to rosserial_server

stertingen:
I tried to mimic the behavior of the python server implementation in this one.

### DIFF
--- a/rosserial_server/include/rosserial_server/session.h
+++ b/rosserial_server/include/rosserial_server/session.h
@@ -501,7 +501,7 @@ private:
         req_param.response.strings.push_back(value);
         break;
       case XmlRpc::XmlRpcValue::TypeDateTime:
-        throw std::runtime_error("Type DateTime is unpussorted.");
+        throw std::runtime_error("Type DateTime is unsupported.");
         break;
       case XmlRpc::XmlRpcValue::TypeBase64:
         throw std::runtime_error("Type Base64 is unsupported.");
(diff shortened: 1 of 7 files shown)


stertingen:
I'm very sorry for the excessive abuse of the CI. I cannot run my tests locally due gtest/gmock issues on arch. :(

(https://github.com/ros/catkin/issues/1030)

stertingen:
It looks like I need some help to get the tests working. (Non-python tests work now.) These work fine using the socket implementation, but the tests on the serial setup time out. (See CI)

I'm sorry that I've not seen PR https://github.com/ros-drivers/rosserial/pull/421 before starting to work on parameters, but since my implementation was also locally tested, I'd let @mikepurvis choose. (Advantage of my implementation: I always return an empty response to the client in case of an error; but that would be easy to fix in @martonmiklos implementation.)

If the other PR gets merged and I got the tests working by then, I'd just provide the tests for the other PR as well.

martonmiklos:
Hi @stertingen 

I got to the same state: I have not managed to get the native tests working so this is why that PR got bitrotten.  Anyway it would be great to get either of the two solution to get upstreamed. I just updated to Ubuntu 20.04 and I have some troubles getting ROS working, however I will take a look on it again once I got everything rolling.

acxz:
@stertingen just want to point out that `ros-noetic-catkin` has been patched and released on Arch Linux such that now the tests work.

PR #486, opened 2020-04-15
Python 3 compatibility for generate_client_ros_lib (test)

stertingen:


mikepurvis:
These fixes are incorporated into #480, targeting `noetic-devel`.

Issue #488, opened 2020-04-16
serial_node.py with use_sim_time issue


daisukes:
The main loop of serial_node.py waits 1 millisecond with `rospy.sleep(0.001)` after receiving a message from serial [here](https://github.com/ros-drivers/rosserial/blob/noetic-devel/rosserial_python/src/rosserial_python/SerialClient.py#L544).

This is okay without simulation, but gazebo_ros publishes clock in 100 Hz and let serial_noe.py wait 10 milliseconds after each message. 

My Arduino publishes 4 different messages in 50 Hz (total 200 Hz) and got sync error when I run it with gazebo (use_sim_time=true).

Is it possible to change it to `time.sleep(0.001)`?

mikepurvis:
That sounds reasonable— Gazebo would have no control over whatever is on the other end of a rosserial connection, so there'd be no reason to alter its view of time. Please send a PR.

PR #489, opened 2020-04-16
use time.sleep instead of rospy.sleep. 

daisukes:
https://github.com/ros-drivers/rosserial/issues/488

Signed-off-by: Daisuke Sato <daisukes@cmu.edu>


mikepurvis:
Looks good, thanks. Closes #488.

Issue #450, opened 2019-11-18
Can not find rosserial_arduino from PyImport_ImportModule 


tongtybj:
When I tried [SevriveClient](https://github.com/ros-drivers/rosserial/tree/melodic-devel/rosserial_arduino/src/ros_lib/examples/ServiceClient) with rosserial_server, I got following warning:

```
chou@chou-ThinkPad-T480:~/ros/rosserial_ws/src/rosserial/rosserial_arduino/src/rosserial_arduino$ roslaunch rosserial_server  serial.launch port:=/dev/ttyDUMMY --screen -v 
... logging to /home/chou/.ros/log/df6e69d0-09fa-11ea-83f4-04ea56a8c113/roslaunch-chou-ThinkPad-T480-10634.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

... loading XML file [/opt/ros/kinetic/etc/ros/roscore.xml]
... executing command param [rosversion roslaunch]
Added parameter [/rosversion]
... executing command param [rosversion -d]
Added parameter [/rosdistro]
Added core node of type [rosout/rosout] in namespace [/]
... loading XML file [/home/chou/ros/rosserial_ws/src/rosserial/rosserial_server/launch/serial.launch]
Added parameter [/rosserial_server/port]
Added node of type [rosserial_server/serial_node] in namespace [/]
started roslaunch server http://chou-ThinkPad-T480:44421/

SUMMARY
========

PARAMETERS
 * /rosdistro: kinetic
 * /rosserial_server/port: /dev/ttyDUMMY
 * /rosversion: 1.12.14

NODES
  /
    rosserial_server (rosserial_server/serial_node)

auto-starting new master
process[master]: started with pid [10648]
ROS_MASTER_URI=http://localhost:11311

setting /run_id to df6e69d0-09fa-11ea-83f4-04ea56a8c113
process[rosout-1]: started with pid [10661]
started core service [/rosout]
process[rosserial_server-2]: started with pid [10668]
[ INFO] [1574078369.186797626]: rosserial_server session configured for /dev/ttyDUMMY at 57600bps.
[ INFO] [1574078369.186875598]: Opened /dev/ttyDUMMY
[ WARN] [1574078369.358219379]: Unable to look up service definition: Unable to import message module rosserial_arduino.
[...]

PR #491, opened 2020-04-17
Only initialize embedded python interpreter once.

mikepurvis:
Should address #450.

@tongtybj, @TobinHall, @stertingen Can you give this a try and confirm that it addresses the issue you were seeing with `PyImport_ImportModule`?

### DIFF
--- a/rosserial_server/src/msg_lookup.cpp
+++ b/rosserial_server/src/msg_lookup.cpp
@@ -35,6 +35,11 @@ namespace rosserial_server
 
 const MsgInfo lookupMessage(const std::string& message_type, const std::string submodule)
 {
+  // Lazy-initialize the embedded Python interpreter. Avoid calling the corresponding
+  // finalize method due to issues with importing cyaml in the second instance. The
+  // total memory cost of having this in-process is only about 5-6MB.
+  Py_Initialize();
+
   MsgInfo msginfo;
   size_t slash_pos = message_type.find('/');
   if (slash_pos == std::string::npos)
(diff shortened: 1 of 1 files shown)


rosserial_server/src/msg_lookup.cpp
```diff
@@ -35,6 +35,14 @@ namespace rosserial_server
 
 const MsgInfo lookupMessage(const std::string& message_type, const std::string submodule)
 {
+  // Lazy-initialize the embedded Python interpreter. The total
+  // memory cost of having this in process is only about 5-6MB.
+  static bool py_initialized = false;
+  if (!py_initialized)
```
stertingen: This check is not really needed. Calling `Py_Initialize()` a second time is considered a no-op, if we really do want to check, we could use `Py_IsInitialized()`.
```suggestion
  if (!Py_IsInitialized())
```

https://docs.python.org/3.5/c-api/init.html#c.Py_IsInitialized

This could possibly be still messed up if `Py_Finalize()` is called somewhere else.

mikepurvis: > Calling `Py_Initialize()` a second time is considered a no-op

~~Is this the case? I'm unable to find documentation supporting this.~~

Never mind that, right here: https://docs.python.org/3/c-api/init.html#c.Py_Initialize

>  This is a no-op when called for a second time (without calling `Py_FinalizeEx()` first).

Will change.

mikepurvis:
Would be interesting to see if the `ast` module can be repeatedly imported— I suspect so, and that would side-step a lot of this, if we only *parsed* the message file rather than actually importing it. Main drag would be having to have more of the actual logic implemented in Python (using `importlib` to located the correct source file, `ast` to parse it, and then navigating the parsed tree to extract the two strings we care about). So we'd either have to have that be a script located somewhere that the embedded interpreter could invoke, or have it be inlined in the cpp file as a string. Neither seems ideal.

Anyway, a project for another day if someone wants to look. :)

PR #494, opened 2020-04-17
Do not fail when trying to update ros_lib which already exists

stertingen:
When running `rosrun rosserial_client make_libraries` somewhere where the `ros_lib` directory already exists (for updating message definitions), the script terminates with an error message.

I decided to add these checks in the same way as in line 546, for the sake of consistency. (Python 3.2 and later allow an additional parameter `exist_ok` for this, but that could not be backported to Melodic.)


mikepurvis:
Looks good, thanks. Closes #454.

mikepurvis:
Oh, didn't see that this was targeting `melodic-devel` already. Picked to `noetic-devel` in e08059e76c4b8d558e616a30e5c3aa5911b4a92f.

PR #497, opened 2020-04-28
Fix/time

rikba:
In https://github.com/ros-drivers/rosserial/issues/169 @PaulBouchier mentions that the implementation of subtraction and addition of time and duration may be erroneous. We actually observed that 
```
ros::Time time = {1, 0};
time -= ros::Duration(0, 1000000);
```
indeed lead to a wrong result on our system.

This PR implements the fix provided by @chuck-h and additional unit tests. 

~~Since I cannot run the unit tests on my machine I'm creating this draft PR first.~~

### DIFF
--- a/rosserial_client/CMakeLists.txt
+++ b/rosserial_client/CMakeLists.txt
@@ -8,12 +8,15 @@ catkin_package(
 
 catkin_python_setup()
 
+add_library(${PROJECT_NAME} src/ros_lib/duration.cpp src/ros_lib/time.cpp)
+
 if(CATKIN_ENABLE_TESTING)
   find_package(rosserial_msgs REQUIRED)
   include_directories(src ${rosserial_msgs_INCLUDE_DIRS})
   catkin_add_gtest(float64_test test/float64_test.cpp)
   catkin_add_gtest(subscriber_test test/subscriber_test.cpp)
   catkin_add_gtest(time_test test/time_test.cpp)
+  target_link_libraries(time_test ${PROJECT_NAME})
 endif()
 
 install(
--- a/rosserial_client/CMakeLists.txt
+++ b/rosserial_client/CMakeLists.txt
@@ -8,15 +8,12 @@ catkin_package(
 
 catkin_python_setup()
 
-add_library(${PROJECT_NAME} src/ros_lib/duration.cpp src/ros_lib/time.cpp)
-
 if(CATKIN_ENABLE_TESTING)
   find_package(rosserial_msgs REQUIRED)
   include_directories(src/ros_lib ${rosserial_msgs_INCLUDE_DIRS})
   catkin_add_gtest(float64_test test/float64_test.cpp)
   catkin_add_gtest(subscriber_test test/subscriber_test.cpp)
-  catkin_add_gtest(time_test test/time_test.cpp)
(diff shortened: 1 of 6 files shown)


rikba:
Ok, I was hoping Travis has a different workspace setup than I do, but I get the same redefinition error of `ros::Duration` here.

Can someone help me? How do I [#include "ros_lib/ros/duration.h"](https://github.com/ros-drivers/rosserial/pull/497/files#diff-beb6d5286300e301d7444ade1c0317ccR1) in the unit test without redefining `ros::Duration`?

```
In file included from /home/travis/catkin_ws/src/rosserial/rosserial_client/test/time_test.cpp:2:

In file included from /home/travis/catkin_ws/src/rosserial/rosserial_client/src/ros_lib/ros/time.h:38:

/opt/ros/melodic/include/ros/duration.h:108:20: error: redefinition of

      'Duration'

class ROSTIME_DECL Duration : public DurationBase<Duration>

                   ^

/home/travis/catkin_ws/src/rosserial/rosserial_client/src/ros_lib/ros/duration.h:46:7: note: 

      previous definition is here

class Duration

      ^

[ 96%] Built target rosserial_test_publish_subscribe

1 error generated.

rosserial/rosserial_client/CMakeFiles/time_test.dir/build.make:62: recipe for target 'rosserial/rosserial_client/CMakeFiles/time_test.dir/test/time_test.cpp.o' failed

make[3]: *** [rosserial/rosserial_client/CMakeFiles/time_test.dir/test/time_test.cpp.o] Error 1

CMakeFiles/Makefile2:2431: recipe for target 'rosserial/rosserial_client/CMakeFiles/time_test.dir/all' failed

make[2]: *** [rosserial/rosserial_client/CMakeFiles/time_test.dir/all] Error 2

CMakeFiles/Makefile2:77: recipe for target 'CMakeFiles/tests.dir/rule' failed

make[1]: *** [CMakeFiles/tests.dir/rule] Error 2

Makefile:175: recipe for target 'tests' failed

make: *** [tests] Error 2

Invoking "make tests -j2 -l2" failed

The command "catkin_make tests" exited with 1.
```

rosserial_client/CMakeLists.txt
```diff
@@ -10,9 +10,10 @@ catkin_python_setup()
 
 if(CATKIN_ENABLE_TESTING)
   find_package(rosserial_msgs REQUIRED)
-  include_directories(src ${rosserial_msgs_INCLUDE_DIRS})
+  include_directories(src/ros_lib ${rosserial_msgs_INCLUDE_DIRS})
```
mikepurvis, line 13: I think this is fine for now, but given the potential for weird header cross talk issues here between tests, we should probably move this over to use `target_include_directories` on a per-test basis to ensure that each one gets just what it needs.

mikepurvis:
> How do I #include "ros_lib/ros/duration.h" in the unit test without redefining `ros::Duration`?

You figured out one way. Another is to wrap the include in a namespace to isolate it from its surroundings; we do this for the `rosserial_test` node, as it is a single binary that is simultaneously a rosserial client and a roscpp node:

https://github.com/ros-drivers/rosserial/blob/noetic-devel/rosserial_test/src/publish_subscribe.cpp#L4-L7

PR #498, opened 2020-05-05
Method to compute duration between time stamps

rikba:
This PR adds a method to compute the difference between two time stamps `Duration operator -(const Time &rhs)`.

We use this for example to compute the duration between two strobe signals of a camera or to compute durations between reading sensor signals.


stertingen:
I think, this should be added to the noetic release as it is consistent with the roscpp behavior: https://wiki.ros.org/roscpp/Overview/Time

mikepurvis:
Looks great, and thanks for also updating the test.

PR #500, opened 2020-05-27
Updates for python3 compatibility

theseankelly:
Came up with these changes while hacking through enabling ROS Noetic on my Turtlebot.

* use python3 import syntax in init
* Use io.BytesIO instead of StringIO
* Queue package renamed queue in python3
* Work with byte arrays instead of unicode strings


theseankelly:
It's come to my attention that these changes already are part of #480  -- it looks like I've missed a couple things, and #480 has already gone through code review. So, I am closing this PR

PR #501, opened 2020-06-10
output_path missing "/" for "copy ros_lib stuff"

elXaptation:
Not exactly sure what happened in https://github.com/ros-drivers/rosserial/pull/495/commits/c7cd984e1dc53e64b7f9323c09a1a7c7c2c09416 but I needed to add a "/" to line-80 "output_path = os.path.join(sys.argv[1], "ros_lib/")" so make_libraries.py will copy and find files in to the Arduino/libraries/ros_lib directory.

### DIFF
--- a/rosserial_arduino/src/rosserial_arduino/make_libraries.py
+++ b/rosserial_arduino/src/rosserial_arduino/make_libraries.py
@@ -77,7 +77,7 @@ if (len(sys.argv) < 2):
 
 # get output path
 path = sys.argv[1]
-output_path = os.path.join(sys.argv[1], "ros_lib")
+output_path = os.path.join(sys.argv[1], "ros_lib/")
 print("\nExporting to %s" % output_path)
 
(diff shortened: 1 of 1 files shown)


mikepurvis:
@stertingen I know we mass-updated this in #495. Can you revisit and see if it needs to be updated everywhere to include the trailing slash? Or perhaps update `rosserial_client_copy_files` so that it does the right thing regardless of the trailing slash being there or not?

stertingen:
I cannot reproduce this behavior. Looking at `rosserial_client_copy_files` I wonder how this can go wrong, because the path (the one that appears to miss a trailing `/`) that is passed into the function is concatenated with subdirectories again, also using `os.path.join()`. Unless there is a leading `/` in a path component, `os.path.join` just eliminates intermediate slashes.

@elXaptation What python version do you use?

elXaptation:
I'm using Python 2.7.17 for the work related to ros-drivers/rosserial

PR #508, opened 2020-08-20
Feature/noetic python support

asukiaaa:
The commits are cherry picked by theseankelly to [noetic-turtlebot branch](https://github.com/theseankelly/rosserial/tree/noetic-turtlebot) branch from [gcc7-fixes](https://github.com/ros-drivers/rosserial/tree/gcc7-fixes) branch.

### DIFF
--- a/rosserial_client/src/ros_lib/ros/msg.h
+++ b/rosserial_client/src/ros_lib/ros/msg.h
@@ -53,7 +53,7 @@ public:
   /**
    * @brief This tricky function handles promoting a 32bit float to a 64bit
    *        double, so that AVR can publish messages containing float64
-   *        fields, despite AVV having no native support for double.
+   *        fields, despite AVR having no native support for double.
    *
    * @param[out] outbuffer pointer for buffer to serialize to.
    * @param[in] f value to serialize.
@@ -63,7 +63,10 @@ public:
    */
   static int serializeAvrFloat64(unsigned char* outbuffer, const float f)
   {
-    const int32_t* val = (int32_t*) &f;
+    // Cast through void* to silence a GCC warning about this being a
+    // non-portable cast (we don't care, it only has to work on AVR).
+    const void* vval = reinterpret_cast<const void*>(&f);
+    const int32_t* val = reinterpret_cast<const int32_t*>(vval);
     int16_t exp = ((*val >> 23) & 255);
     uint32_t mantissa = *val & 0x7FFFFF;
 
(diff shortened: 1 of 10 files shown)


mikepurvis:
I guess this is still an omnibus in the way that #480 was, but at this point pragmatism should carry the day. Please also re-enable the python_socket test here, and if that passes clean on Noetic CI, then we're good to go:

https://github.com/ros-drivers/rosserial/blob/04ddab73643e617910cbfdfe6dcf60476385bffe/rosserial_test/CMakeLists.txt#L47

asukiaaa:
Thank you for the information.
I merged your commit to my branch [asukiaaa:feature/noetic-python-support](https://github.com/asukiaaa/rosserial/tree/feature/noetic-python-support) to activate test.
Could you create PR to my(asukiaaa's) branch not to mikepurvis's branch on next?

rosserial_client/src/ros_lib/ros/msg.h
```diff
@@ -63,7 +63,10 @@ class Msg
    */
   static int serializeAvrFloat64(unsigned char* outbuffer, const float f)
   {
-    const int32_t* val = (int32_t*) &f;
+    // Cast through void* to silence a GCC warning about this being a
+    // non-portable cast (we don't care, it only has to work on AVR).
+    const void* vval = reinterpret_cast<const void*>(&f);
+    const int32_t* val = reinterpret_cast<const int32_t*>(vval);
```
stertingen, line 69: I think the portable way not invoking undefined behavior is `std::memcpy`. Most compiler optimize this anyway as memcpy usually is a compiler builtin.

From cppreference.com (https://en.cppreference.com/w/cpp/string/byte/memcpy):
> Where strict aliasing prohibits examining the same memory as values of two different types, std::memcpy may be used to convert the values.

rosserial_client/src/ros_lib/ros/msg.h
```diff
@@ -122,10 +125,12 @@ class Msg
...
   static int deserializeAvrFloat64(const unsigned char* inbuffer, float* f)
   {
-    uint32_t* val = (uint32_t*)f;
     int16_t exp;
     uint32_t mantissa;
-
+    // Cast through void* to silence a GCC warning about this being a
+    // non-portable cast (we don't care, it only has to work on AVR).
+    void* vval = reinterpret_cast<void*>(f);
+    uint32_t* val = reinterpret_cast<uint32_t*>(vval);
```
stertingen, line 133: Also `std::memcpy` here.

stertingen (commented):
The PR currently silences a warning about undefined behavior without actually fixing it. The portable way is `std::memcpy`.

rosserial_client/src/ros_lib/ros/msg.h
mikepurvis: Sounds good. I'm going to merge this now, as it's passing, and then put up a separate PR to reform this.

asukiaaa:
Thank you for checking and merging.

mikepurvis:
Thank _you_ for the work rebasing this branch and putting up the PR.

Issue #499, opened 2020-05-26
Noetic release


doisyg:
Hi,
Is there something blocking it?

mikepurvis:
Both `rosserial_python` tests are currently disabled:

https://github.com/ros-drivers/rosserial/blob/04ddab73643e617910cbfdfe6dcf60476385bffe/rosserial_test/CMakeLists.txt#L47-L49

I believe it can be got going again with the `rosserial_python` changes extracted from #480 (that PR had become a bit of an omnibus):

https://github.com/ros-drivers/rosserial/pull/480/files#diff-2d3c0dc20802b1a68f2ff2955a5037ed

I (or someone) needs to get back to it so our initial Noetic version isn't completely broken. :)

theseankelly:
I submitted PR #500 yesterday based on some weekend hacking. Totally missed this issue and the aforementioned PR #480 

I'm not a python expert but my PR seems to get the trick done for rosserial on TurtleBot3 for OpenCR.

theseankelly:
I ended up closing my PR (#500). My changes missed a few things which are already in #480. Plus, #480 has already gone through code review. I think the best path is to rebase and merge #480.

BriceRenaudeau:
There is still Python compatibility errors in PR (#500 and #480) :
in `class ServiceServer:` in `def callback(self, req):`
Line 171(in PR 500)
```
if self.parent.send(self.id, data_buffer.getvalue()) >= 0:
            while self.response is None:
                pass
```
This makes fatal crashes when calling service on the device.
-> NoneType can not be compared with Int!

Need to be replaced by :
```
self.parent.send(self.id, data_buffer.getvalue())
        while self.response is None:
            pass
```
This change was tested on real robot.

Do I need to make a PR with all python 3 compatibilities like @theseankelly (thanks for pre-clean) or wait for the modification to be included in PR #480?

asukiaaa:
I could run rosserial on noetic with using [gcc7-fixes](https://github.com/ros-drivers/rosserial/tree/gcc7-fixes) branch that was created for PR https://github.com/ros-drivers/rosserial/pull/480.

@mikepurvis 
Thank you for updating codes.
When will the PR https://github.com/ros-drivers/rosserial/pull/480 be merged?

mikepurvis:
[...]

PR #514, opened 2020-09-07
Fix python3

BriceRenaudeau:
Following Issue #499


mikepurvis:
Great, thanks. I don't personally use `rosserial_python`, so I depend on the tests and users to catch issues with it.

PR #516, opened 2020-09-10
Fixing message size issue.

TobinHall:
In rosserial_server the message is published with the rosserial checksum byte.
This means that subscribers receive an extra byte in each message.
Some clients (like [PlotJuggler)](https://github.com/facontidavide/PlotJuggler) consider the size mismatch an error.


mikepurvis:
Nice catch! Thanks for the fix.

PR #517, opened 2020-09-10
Fixing UDP client socket used for receiving data.

TobinHall:
The UDP implementation seems to receive on the client endpoint rather than the server endpoint.


TobinHall:
Upon further testing, this is not necessary.

PR #522, opened 2020-09-24
Remove "override" mark form "ServiceClient::call()" function

tongtybj:
`ServiceClient::call ()` does not have a virtual function in ` Subscriber_`, so remove the `override` mark.

### DIFF
--- a/rosserial_client/src/ros_lib/ros/service_client.h
+++ b/rosserial_client/src/ros_lib/ros/service_client.h
@@ -54,7 +54,7 @@ public:
     this->waiting = true;
   }
 
-  virtual void call(const MReq & request, MRes & response) override
+  virtual void call(const MReq & request, MRes & response)
   {
     if (!pub.nh_->connected()) return;
     ret = &response;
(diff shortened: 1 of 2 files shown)


tongtybj:
Also exclude  the`ServiceClient` instance in unregistration process, otherwise following error will occur:

```
  File "/root/ros/rosserial_ws/devel/lib/rosserial_python/serial_node.py", line 15, in <module>
    exec(compile(fh.read(), python_script, 'exec'), context)
  File "/root/ros/rosserial_ws/src/rosserial/rosserial_python/nodes/serial_node.py", line 72, in <module>
    server.listen()
  File "/root/ros/rosserial_ws/src/rosserial/rosserial_python/src/rosserial_python/SerialClient.py", line 251, in listen
    self.startSerialClient()
  File "/root/ros/rosserial_ws/src/rosserial/rosserial_python/src/rosserial_python/SerialClient.py", line 272, in startSerialClient
    srv.unregister()
AttributeError: 'ServiceClient' object has no attribute 'unregister'

```

Issue #518, opened 2020-09-11
cstring: No such file or directory


shmulike:
Hello
I'm using Arduino IDE 1.8.13 and Arduino Uno
When I try to verify/compile the HelloWorld example I'm getting this error in the IDE:

```
In file included from /home/user/Arduino/libraries/ros_lib/std_msgs/Time.h:7:0,
                 from /home/user/Arduino/libraries/ros_lib/ros/node_handle.h:40,
                 from /home/user/Arduino/libraries/ros_lib/ros.h:38,
                 from /home/user/Arduino/libraries/ros_lib/examples/HelloWorld/HelloWorld.pde:6:
/home/user/Arduino/libraries/ros_lib/ros/msg.h:40:10: fatal error: cstring: No such file or directory
 #include <cstring>
          ^~~~~~~~~
compilation terminated.
exit status 1
Error compiling for board Arduino Uno.
```

jagyer:
I have exactly the same problem! Exactly!
Arduino IDE 1.8.14, Ubuntu 20.04 LTS, ROS Noetic.
Installing rosserial-arduino using binaries just began to work a few days ago (as described in 2.1.1): 
http://wiki.ros.org/rosserial_arduino/Tutorials/Arduino%20IDE%20Setup
Seems like it at least, no errors. And now this, it's so frustrating...
Any advice much appreciated!
Would it help to post the verbose compilation output message from Arduino IDE? It contains a lot...

shmulike:
Hi
Still didn't find any solution for that

dmitriypoyarkov:
Me too, waiting if there will be somebody to solve this.
In file included from /home/fc/Arduino/libraries/ros_lib/std_msgs/Time.h:7:0,
                 from /home/fc/Arduino/libraries/ros_lib/ros/node_handle.h:40,
                 from /home/fc/Arduino/libraries/ros_lib/ros.h:38,
                 from /home/fc/catkin_ws/src/common_canonical/led_strip_control/firmware/ledstrip_ros/ledstrip_ros.ino:2:
/home/fc/Arduino/libraries/ros_lib/ros/msg.h:40:10: fatal error: cstring: No such file or directory
 #include <cstring>
          ^~~~~~~~~
compilation terminated.

RuslanNig:
the same problem

fadamsyah:
Hi
I have the same issue

Guasch5000:
Maybe this can help a bit.

[...]

PR #525, opened 2020-10-01
Changed C++ std::memcpy() into C version

rafaol:
The `cstring` include and `std::memcpy` code were causing issues with rosserial_arduino #518 . Arduino by default does not have the C++ standard library available, and "ros/msg.h" is copied by rosserial_arduino's "make_libraries.py" into the generated ros_lib Arduino library. The fix was simply to replace `memcpy` by its equivalent C version.

### DIFF
--- a/rosserial_client/src/ros_lib/ros/msg.h
+++ b/rosserial_client/src/ros_lib/ros/msg.h
@@ -37,7 +37,7 @@
 
 #include <stdint.h>
 #include <stddef.h>
-#include <cstring>
+#include <string.h>
 
 namespace ros
 {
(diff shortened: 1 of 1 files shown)


RBinsonB:
Any update on this merge request? Using this branch was the only way I could resolve issue #518.

Samu24:
maybe @mikeferguson @mikepurvis

mikeferguson:
> maybe @mikeferguson @mikepurvis

I'm no longer active on this project.

mikepurvis:
I'm surprised this is needed, as the target for rosserial has always been a C++ compiler where the basic stdlib _should_ be available. However, it looks like a lot of people are affected in the other thread, so it makes sense to fix it for them.

PR #530, opened 2020-10-30
Use shutil.copyfile() instead of shutil.copy()

lopsided98:
On my distribution, rosserial's files are installed readonly, and `shutil.copy()` preserves this when copying them. The causes `make_libraries` to fail the next time it runs as it tries to overwrite the readonly files. `shutil.copyfile()` does not preserve the permissions, preventing this problem.


mikepurvis:
Awesome, thanks for the easy fix.

PR #539, opened 2021-01-17
support for ESP32 ethernet

niktwo17:
Adds support for ESP32 boards using ethernet such as the Olimex ESP32-EVB.

### DIFF
--- a/rosserial_arduino/src/ros_lib/ArduinoTcpHardware.h
+++ b/rosserial_arduino/src/ros_lib/ArduinoTcpHardware.h
@@ -40,6 +40,8 @@
   #include <ESP8266WiFi.h>
 #elif defined(ESP32)
   #include <WiFi.h> // Using Espressif's WiFi.h
+#elif defined(ESP32_ETH)
+  #include <ETH.h> //using Espressif's WiFi.h ETH option
 #else
   #include <SPI.h>
   #include <Ethernet.h>
@@ -59,7 +61,7 @@ public:
 
   IPAddress getLocalIP()
   {
-#if defined(ESP8266) or defined(ESP32)
+#if defined(ESP8266) or defined(ESP32) or defined(ESP32_ETH)
     return tcp_.localIP();
 #else
     return Ethernet.localIP();
@@ -104,7 +106,7 @@ public:
   }
 
 protected:
-#if defined(ESP8266) or defined(ESP32)
+#if defined(ESP8266) or defined(ESP32) or defined(ESP32_ETH)
   WiFiClient tcp_;
 #else
   EthernetClient tcp_;
--- a/rosserial_arduino/src/ros_lib/ros.h
+++ b/rosserial_arduino/src/ros_lib/ros.h
@@ -37,7 +37,7 @@
 
 #include "ros/node_handle.h"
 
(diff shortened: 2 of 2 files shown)


romainreignier:
Hi @niktwo17

Thank you for your contribution.

I have tried to use Ethernet on Olimex ESP32-EVB board without your patch on `melodic-devel` branch and it already works.
So I am wondering why did you need this patch?
First of all, I did not find the `ESP32_ETH` definition in the [arduino-esp32](https://github.com/espressif/arduino-esp32) repo.
Where does it come from?

To test it, I have used this sketch adapted from [TcpBlink](https://github.com/ros-drivers/rosserial/tree/noetic-devel/rosserial_arduino/src/ros_lib/examples/TcpBlink):
```c++
/*
 * rosserial Subscriber Example using Ethernet on ESP32
 * Blinks an LED on callback
 */

#include <ETH.h>

#include <ros.h>
#include <std_msgs/Empty.h>

ros::NodeHandle  nh;

// Shield settings
IPAddress ip(192, 168, 10, 177);
IPAddress gateway(192, 168, 10, 1);
IPAddress subnet(255, 255, 0, 0);

// Server settings
IPAddress server(192, 168, 10, 24);
uint16_t serverPort = 11411;

const uint8_t ledPin = 12;

void messageCb( const std_msgs::Empty&){
  digitalWrite(ledPin, HIGH-digitalRead(ledPin));   // blink the led
}

ros::Subscriber<std_msgs::Empty> sub("toggle_led", &messageCb );

static bool eth_connected = false;

void WiFiEvent(WiFiEvent_t event)
{
  switch (event) {
    case SYSTEM_EVENT_ETH_START:
      Serial.println("ETH Started");
      //set eth hostname here
      ETH.setHostname("esp32-ethernet");
      break;
    case SYSTEM_EVENT_ETH_CONNECTED:
      Serial.println("ETH Connected");
      break;
    case SYSTEM_EVENT_ETH_GOT_IP:
      Serial.print("ETH MAC: ");
      Serial.print(ETH.macAddress());
      Serial.print(", IPv4: ");
      Serial.print(ETH.localIP());
      if (ETH.fullDuplex()) {
        Serial.print(", FULL_DUPLEX");
      }
      Serial.print(", ");
      Serial.print(ETH.linkSpeed());
      Serial.println("Mbps");
      eth_connected = true;
      break;
    case SYSTEM_EVENT_ETH_DISCONNECTED:
      Serial.println("ETH Disconnected");
      eth_connected = false;
      break;
    case SYSTEM_EVENT_ETH_STOP:
      Serial.println("ETH Stopped");
      eth_connected = false;
      break;
    default:
      break;
  }
}

void setup()
{
  Serial.begin(115200);
  WiFi.onEvent(WiFiEvent);
  ETH.begin();
  ETH.config(ip, gateway, subnet);
  
  pinMode(ledPin, OUTPUT);
  
  nh.getHardware()->setConnection(server, serverPort);
  nh.initNode();
  nh.subscribe(sub);
}

void loop()
{
  nh.spinOnce();
  delay(1);
}
```

niktwo17:
Hi, thanks for coming back and checking so fast.

Using the ETH.h gave me quite a headache for some time, so I ended up adapting the arduinotcphardware file to make it work.
The definition was derived from my project, so in this context not useful I have to admit, your example is more efficient and consistent.

Would be glad if you could add an TcpBlink_ESP32  example then.

PR #543, opened 2021-01-22
Add separate namespace for log messages from clients (#2)

brawner:
I was curious what you all thought about adding a separate namespace for the log messages coming from clients? For my own project, I want to be able to set the logger level specifically for the client without confusing it with the other log messages coming from rosserial_server.

It doesn't look like there is a simple way to attach a client-specific name for this named logger, so this just adds one namespace for all clients.

Signed-off-by: Stephen Brawner <brawner@gmail.com>


mikepurvis:
This seems totally reasonable to me, though obviously it is a behaviour change. We no longer have a ROS 1 major release coming up upon which to hang breaking changes, so I'm not sure how disruptive people would find this kind of thing.

The safest thing would be to pass the client log namespace as a parameter which defaults to `""`, so that the new behaviour is opt-in.

PR #521, opened 2020-09-24
[rosserial_arduino] correct module import rule for Python3

tongtybj:
Related to  #519 #504

Issue #519, opened 2020-09-23
Can not export msg and srv in rosserial_arduino by make_libraries.


tongtybj:
When I tried `rosrun rosserial_arduino make_libraries.py ~/` with the latest version of `noetic-devel`,
 I got following unexpected result about `rosserial_arduino` package:
```
Exporting rosserial_arduino

  Messages:
Unable to build message: rosserial_arduino/Adc
    Adc,Exporting rosserial_arduino

  Services:
Unable to build service: rosserial_arduino/Test
    Test,

Exporting rosserial_mbed

  Messages:
    Adc,

  Services:
    Test,

Exporting rosserial_msgs

  Messages:
    TopicInfo,Log,

  Services:
    RequestParam,
```
Then, I traced the commit history, and found that before commit  a7ac6cd5f55401148a8e8c854d9029f3e815d8e5, `rosserial_arduino/Adc.msg` and `rosserial_arduino/Test.srv` can be correctly generated in ros_lib. 
Then I checked which changed file  induce such problem. Surprisingly, when I reverted the change of `rosserial_python/src/rosserial_python/SerialClient.py ` in this commit, the problem disappeared. 

Can anyone else (@mikeodr  @asukiaaa ) try following commands to confirm whether this is a common phenomenon or my personal problem?

`git checkout noetic-devel # latest version`
`rosrun rosserial_arduino make_libraries.py ~/  # please check there is no rosserial_arduino inside ros_lib`

`git checkout a7ac6cd`
`rosrun rosserial_arduino make_libraries.py ~/  # please check there is no rosserial_arduino inside ros_lib`

`git reset HEAD~`
`git checkout rosserial_python/src/rosserial_python/SerialClient.py`
`rosrun rosserial_arduino make_libraries.py ~/  # please check there is  rosserial_arduino inside ros_lib`

mikeodr:
I think you mean @mikepurvis and not myself. 

I am no longer active with ROS presently. Best of luck with your issue.

tongtybj:
@mikeodr 

Thank you for your reply and sorry to bother you.

tongtybj:
@mikepurvis 

I narrow down this problem, and found following related error:

```
root@fb70210db6ac:~/ros/rosserial_ws# python3 -c 'import rosserial_arduino'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
[...]

Issue #504, opened 2020-07-09
rosserial_python serial_node: No module named 'SerialClient'


shmulike:
Hello
I am using Ubuntu Focal 20.04, with Ros Noetic (default python3)
I downloaded the rosserial source package: 'noetic-devel'

I uploaded the example for hello worlds to Arduino Nano.

When I try to run:
`rosrun rosserial_python serial_node.py /dev/ttyUSB0`

I receive that message:
```
Traceback (most recent call last):
  File "/home/me-ubuntu/catkin_ws/devel/lib/rosserial_python/serial_node.py", line 15, in <module>
    exec(compile(fh.read(), python_script, 'exec'), context)
  File "/home/me-ubuntu/catkin_ws/src/rosserial/rosserial_python/nodes/serial_node.py", line 39, in <module>
    from rosserial_python import SerialClient, RosSerialServer
  File "/home/me-ubuntu/catkin_ws/devel/lib/python3/dist-packages/rosserial_python/__init__.py", line 34, in <module>
    exec(__fh.read())
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'SerialClient'
```

Please advise what to do?

mikepurvis:
The port of rosserial_python to py3 is still in PR limbo on #480. Myself or someone will needs to cherry-pick the relevant commits forward to the current noetic branch.

shmulike:
Do you know maybe when the package will be updated?

My only options left is to downgrade to Ubuntu18 and Melodic

ruffner:
I have found that putting a . in front of SerialClient in  ```rosserial_python/src/rosserial_python/__init__.py``` so that it reads:
```from .SerialClient import *``` 
fixes the SerialClient import error

**Comment by shmulike** [comment]
Yes, I did that, this is one of the changes from Puthon2 to Ptuhon3

Now I am facing another problem that I don't know how to solve. I am getting an error of unrecognized type: **_basestringI_.**
There are saw several questions about that, some of them suggested replacing it with **_byte_** or something like that.
None of them was working for me.

I know that ROS is open-source, but rosserial is a very basic package.
I really hope that the package will be updated soon.
Thanks to all of you who are making it happen !!

**Comment by ruffner** [comment]
To fix that problem, you have to turn *basestring* to just *str*. Then the line after that needs an addition of ```encode()```  added to the end of data  so it reads 
```
elif isinstance(data, str):
    self._write(data.encode())
```

[...]

PR #546, opened 2021-01-31
examples/ADC/ADC.pde: rosserial_arduino -> rosserial_mbed

s-t-a-n:
rosserial_arduino/src/ros_lib/examples/ADC/ADC.pde does not compile. It seems that the ADC components in  rosserial_arduino were recently moved to rosserial_mbed. Changing rosserial_arduino to rosserial_mbed does the trick.

Tested with:
* Arduino IDE 1.8.13
* Arduino-CI 1.3

ROSserial headers generated from noetic-devel using:
* python-pyserial 3.5-1
* ros-noetic-roscpp-serialization 0.7.1-1


romainreignier:
Mbed and Arduino are distinct frameworks, there is no point to mix the two.
The [Adc.msg file](https://github.com/ros-drivers/rosserial/blob/dfc147ec379ad6cdbe83e3ed99d44e35ea0ec29b/rosserial_arduino/msg/Adc.msg) has not been changed for 10 years.
The issue is elsewhere.

s-t-a-n:
The problem seems to be in the call to catkin_python_setup() in rosserial_arduino/CMakeLists.txt. Commenting it yields the generation of the rosserial_arduino messages.

I'll have a further look into it.

s-t-a-n:
Ah luckily someone already found out what's bugging: #519 

> This is related to #504, and is fixed by #521 

I can confirm this is fixed by #521

romainreignier:
Thanks for digging into the issue.

PR #550, opened 2021-03-04
Gracefully handle socket server end.

Nurgak:
When stopping the socket server the `accept()` method blocks until SIGTERM, this is not pretty and takes time...

I suggest a 1 second timeout which fails silently continuously if no connection is present and allows to check for the `rospy.is_shutdown()` condition of the while loop to end gracefully and immediately.


mikepurvis:
This looks reasonable, thanks for the fix!

PR #551, opened 2021-03-04
Update SerialClient.py

altineller:
adds a shutdown hook for sending txStopRequest, and comments out the txtStopRequest before the thread_join.


mikepurvis:
Looks reasonable, thanks.

PR #553, opened 2021-03-06
Fix CMP0038 warning on CMake >= 3.0

VasilyF:
As of CMake 3.0 targets may not link to themselves when calling `target_link_libraries` (policy CMP0038), doing so will raise a warning and ignore targets linking to themselves. Trying to build on noetic, this is reported as an error.

Fix reflects PR to source: https://github.com/queezythegreat/arduino-cmake/pull/143


mikepurvis:
Looks reasonable, thanks.

PR #554, opened 2021-03-12
compatibility with Python3 for rosserial_arduino

mcamurri:
This modification is required for Python3 to find the local `SerialClient` module


mikepurvis:
This was an unfortunate miss from the initial Python 3 porting effort, thanks.

PR #556, opened 2021-04-07
node_handle: spinOnce: add two new return codes

alonblade:
Adds two new return codes (backward compatible change):

SPIN_TX_STOP_REQUESTED
SPIN_TIME_RECV

They are returned as the name suggests to differentiate two states in
which currently SPIN_OK would have been returned.

This allows the caller to further use the return, i.e. issuing a log
message (that is always sent irrespective of connected status since ID <
100).

Signed-off-by: Alon Levy <alon.levy@bladeranger.com>

### DIFF
--- a/rosserial_client/src/ros_lib/ros/node_handle.h
+++ b/rosserial_client/src/ros_lib/ros/node_handle.h
@@ -67,6 +67,8 @@ namespace ros
 const int SPIN_OK = 0;
 const int SPIN_ERR = -1;
 const int SPIN_TIMEOUT = -2;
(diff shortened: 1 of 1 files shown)


alonblade:
@mikepurvis Hi, sorry for the noise - I'm not sure who to ask about this PR, so I just picked you since you did the last release (per the commit log)? Is this something you would consider including? Thanks,

Alon

mikepurvis:
Yes, sorry for the delay. My day job no longer actively uses this repo, so my maintenance of it comes in fits and starts.

This looks like a reasonable change, thanks.

PR #559, opened 2021-04-14
Remove ESP flags for ArduinoTcpHardware to use hardware serial

asukiaaa:
I want to use HardwareSerial with ESP32.
Could you consider removing ESP flags to activate ArduinoTcpHardware?

Thanks.

### DIFF
--- a/rosserial_arduino/src/ros_lib/examples/Esp8266HelloWorld/Esp8266HelloWorld.ino
+++ b/rosserial_arduino/src/ros_lib/examples/Esp8266HelloWorld/Esp8266HelloWorld.ino
@@ -9,6 +9,9 @@
  *
  */
 #include <ESP8266WiFi.h>
+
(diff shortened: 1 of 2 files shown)


romainreignier:
If the serial path works without any more changes, this seems ok but the [esp8266 example](https://github.com/ros-drivers/rosserial/blob/noetic-devel/rosserial_arduino/src/ros_lib/examples/Esp8266HelloWorld/Esp8266HelloWorld.ino) should be changed to add the `ROSSERIAL_ARDUINO_TCP` define at the top.

asukiaaa:
Thank you for the feedback.
I checked `ROSSERIAL_ARDUINO_TCP` flag enables tcp on ESP32 and ESP32 can use hardware serial without it.

mikepurvis:
This seems reasonable and non-breaking. Thanks for the change.

PR #562, opened 2021-05-25
Made tcp_port a local parameter

amalpavithran:
The tcp_port parameter being a global parameter prevents the launch of multiple serial nodes on different port numbers using a launch file

### DIFF
--- a/rosserial_python/nodes/serial_node.py
+++ b/rosserial_python/nodes/serial_node.py
@@ -54,10 +54,15 @@ if __name__=="__main__":
     # TIOCM_DTR_str) line, which causes an IOError, when using simulated port
     fix_pyserial_for_test = rospy.get_param('~fix_pyserial_for_test', False)
 
-    tcp_portnum = int(rospy.get_param('~tcp_port', '11411'))
+    if(rospy.has_param('~tcp_port')):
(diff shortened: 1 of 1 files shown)


mikepurvis:
I agree with this in principle, but it's probably also a breaking change. Could you consider implementing this using a fallback where it looks in the private namespace and if the param isn't there it also checks the legacy name?

Forr consistency the same thing should probably done with the fork_server parameter, though I suppose there's little reason that would ever differ between instances.

amalpavithran:
Have made the necessary changes do have a look and let me know if anymore changes are required 😄

mikepurvis:
Looks great, thanks! There's no need for a fallback in the initial `get_param` case since we'll only hit that if we know it exists, and a comment explaining that the construct is there to not break legacy users would be helpful. But if you don't get to that, I'll merge anyway in a bit.

amalpavithran:
Yeah I was also having doubts about the initial `get_param` then felt that I will just leave it in there as it wasn't causing any issues. Have pushed a commit for that now with the comments you have requested. Do modify the comments as I feel I might not have been able to covey what you have requested.

mikepurvis:
Looks perfect, thanks.

PR #594, opened 2022-07-18
[rosserial_python] Use rospy.on_shutdown in SerialClient shutdown

708yamaguchi:
When I use `serial_node.py` from roslaunch, we cannot kill `serial_node.py` by Ctrl-c.

This is because `serial_node.py` does not catch `SystemExit` exception caused by `sys.exit(0)` in `txStopRequest`.
https://github.com/ros-drivers/rosserial/blob/c169ae2173dcfda7cee567d64beae45198459400/rosserial_python/nodes/serial_node.py#L94-L109

In this PR, I used `rospy.on_shutdown` instead of `signal.signal(signal.SIGINT, self.txStopRequest)`, following [`8fb7954` (#508)](https://github.com/ros-drivers/rosserial/pull/508/commits/8fb795456b5ed39aed1d256f757558b5848cebfa) and https://github.com/ros-drivers/rosserial/pull/551
This change enables to kill `serial_node.py` by Ctrl-c.

Thank you very much for your help, @sktometometo

### DIFF
--- a/rosserial_python/src/rosserial_python/SerialClient.py
+++ b/rosserial_python/src/rosserial_python/SerialClient.py
@@ -397,7 +397,11 @@ class SerialClient(object):
         self.requestTopics()
         self.lastsync = rospy.Time.now()
 
-        signal.signal(signal.SIGINT, self.txStopRequest)
+        def shutdown():
+            self.txStopRequest()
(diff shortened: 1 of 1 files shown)


Hommus:
Thanks for this PR @708yamaguchi!

PR #631, opened 2024-04-25
Patch for support Arduino Due Programming Port as Default Option

racarla96:


racarla96:
Patch for support Arduino Due Programming Port as Default Option

Issue #314, opened 2017-09-09
Some messages with nested array types yield incorrect C++ deserialization code


tom-f-oconnell:
A message of type
```
stimuli/Pulse[] pulse_seq
  uint8 pin
  stimuli/Transition[] states
    time t
    stimuli/State s
      uint64 ms_on
      uint64 ms_off
```
will generate code in `ros_lib/stimuli/Pulse.h` that has a `stimuli::Transition` pointer.  Since each element of the dynamically allocated `stimuli::Pulse` array in `ros_lib/stimuli/PulseSeq.h` is first deserialized in the static field `st_<type_name>` (generated in line 221 of `make_library.py`), and the `realloc` will not typically return new addresses after the first allocation (the pointer in the `states` field is no longer `NULL`), the address in the `states` field is the same for each element of `pulse_seq`.  The `pin` field being set correctly.

This should either be documented as a limitation, or support should be added for messages with this structural motif.  The generated Python functions seem to support this specific message type for both serialization and deserialization.

I could do some work towards a PR, if you decide this should be supported, but I could use some pointers (no pun intended) about initial design choices.  I (think) I realize that it would be harder to implement support for these kinds of messages while still avoiding memory issues.

Feel free to clone my repository at `github.com/tom-f-oconnell/stimuli` to get a copy of the offending message types.  `acf142967` (current `HEAD`) will have them if something happens to change in later commits.

P.s. in a quick look through some of the messages from other packages on my machine (i.e. `std_msgs`, etc), I did not see any messages fitting the criteria I described here.  Is it a ROS idiom / rule not to make messages like this?  I could not find documentation of this restriction, and my messages seem to work with the standard tooling, at least in `rospy`.

mikepurvis:
There's nothing that prohibits this type of message, and it is very much supported in roscpp and rospy. There's some discussion of rosserial's Array-related limitations in this wiki page, but it doesn't seem to cover the particular case you've found:

http://wiki.ros.org/rosserial/Overview/Limitations#Arrays

I'd be delighted for that page to be updated, and I would also totally entertain a PR that lifts or revisits the limitation. Because the wire format of rosserial's serialization is identical to roscpp, the only changes required would be in the client. I don't have any particular tips or ideas on this— you'd have to give it some thought and make a proposal.

tom-f-oconnell:
Ok, thanks Mike.  I added a bit to the end of the Wiki page, and I may work on a PR over the next few weeks when I have time.

[...]

PR #633, opened 2025-10-03
[rosserial_client] fix deserialization for nested array

sugikazu75:
This patch enables to receive array of array in microcontroller correctly by fixing deserialization fofr message.

When sending an array of arrays to the microcontroller, accessing it yielded the same value even though it should have sent different values.
This is caused by a deserialization error resulting in the same pointer being referenced.
A similar report can be found at https://github.com/ros-drivers/rosserial/issues/314#issuecomment-1078665866.

I fixed the part that deserializes messages. I deserialized each element composing the array directly to prevent sharing pointers.


sugikazu75:
I made a repository to test this phenominon https://github.com/sugikazu75/rosserial_nested_array

